PR #25, opened 2020-03-06
Provide desktop launcher file

alerque:
This file is easily generated. My own packaging generates it at build time:

```sh
gendesk -f -n --pkgname ezra-project --pkgdesc ezra-project --name "Ezra Project"
```

Also I believe most stock Electron has some facility for creating launchers if none exist.

The advantage of providing it in the upstream project is that this opens the door to more features:

* A desktop icon can be specified.
* The launcher itself can have localized strings for different languages.

This will also add a little bit of consistency across distros/package systems.

I don't know what needs to be done to tell your other electron packaging to use this, but for my own purposes packaging on Arch just including the source file in the repo somewhere is enough.

### DIFF
--- a/ezra-project.desktop
+++ b/ezra-project.desktop
@@ -1,0 +1,11 @@
+[Desktop Entry]
+Version=1.0
+Type=Application
+Name=Ezra Project
+GenericName=Bible Study Tool
+Comment=Bible study tool focussing on topical study based on keywords/tags
+Exec=ezra-project
+Icon=ezra-project
+Terminal=false
+StartupNotify=false
+Categories=Education;Spirituality;Literature;
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ package-lock.json
 strongs_todo.txt
 .DS_Store
 complexity_metrics.html
+tags
--- a/package_config/deb_config_debian10.json
+++ b/package_config/deb_config_debian10.json
@@ -2,13 +2,13 @@
   "dest": "release/packages/",
   "categories": [
     "Education",
-    "Utility",
-    "Spirituality"
+    "Spirituality",
+    "Literature"
   ],
   "name": "ezra-project",
   "productName": "Ezra Project",
   "genericName": "Bible Study Tool",
-  "description": "Ezra Project is a Bible study tool focussing on topical study based on keywords/tags.",
+  "description": "Ezra Project is a Bible study tool focussing on topical study based on keywords/tags",
   "productDescription": "Ezra Project is a Bible study tool focussing on topical study based on keywords/tags. This program helps the user to easily create and manage topical verse lists. Ezra Project works with SWORD Bible translation modules and thus enables Bible study in many languages.",
   "depends": [
     "libicu63",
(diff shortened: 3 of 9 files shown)


ezra-project.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
+Version=1.0
```
tobias-klein: Should this be the actual Ezra Project version?

ezra-project.desktop
```diff
@@ -0,0 +1,10 @@
...
+Type=Application
+Name=Ezra Project
+Comment=ezra-project
+Exec=ezra-project
+Icon=ezra-project
+Terminal=false
+StartupNotify=false
+Categories=Application;
```
tobias-klein: In the [other package configuration](https://github.com/tobias-klein/ezra-project/tree/master/package_config) files I've been using the following categories:
```
  "categories": [
    "Education",
    "Utility",
    "Spirituality"
  ],
```
Shouldn't we use the same here?

ezra-project.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
+Version=1.0
+Type=Application
+Name=Ezra Project
+Comment=ezra-project
+Exec=ezra-project
+Icon=ezra-project
```
tobias-klein: Good reminder that we still need a nice icon for Ezra Project! :)

ezra-project.desktop
alerque: No, this is the version of the desktop entry specification file format. It affects what fields are required, how categories are parsed, etc.

ezra-project.desktop
alerque: Yes, probably, but we should find a map of categories because those sound like Ubuntu app store categories rather that FSF desktop categories.

ezra-project.desktop
alerque: See #26.

ezra-project.desktop
tobias-klein: These categories are actually the ones from freedesktop.org.

Education and Utility from the [main categories](https://specifications.freedesktop.org/menu-spec/latest/apa.html) and Spirituality from the [additional categories](https://specifications.freedesktop.org/menu-spec/latest/apas02.html) on specifications.freedesktop.org.

I'm using these categories consistently both for *.deb packages and *.rpm packages.

alerque: Here is the 1.0 spec categories: https://specifications.freedesktop.org/menu-spec/menu-spec-1.0.html#category-registry

* "Spirituality" is not a thing.
* Your use of "Utility" is probably incorrect, this is not a utility in the software sense.
* I think "Education" is the only main category that makes sense.
* For secondary categories "DataVisualization" and "Liturature" might make sense.

ezra-project.desktop
alerque: Spec details here: https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-1.1.html#basic-format

ezra-project.desktop
alerque: Oops, my web page search foo failed _and_ I missed your comment because I was typing mine. I'll update....

alerque:
@tobias-klein Sure.

alerque:
Done. I also rebased to cleanup the git history, this was turning into a mess of commits for a small change.

PR #46, opened 2020-04-10
Add simple icon

tobias-klein:
There's several new files under `icons/` now:
* ezra.svg (the SVG icon source, created in Inkscape)
* ezra.png (Icon for Linux)
* ezra.ico (Icon for Windows)
* ezra.icns (Icon for macOS)

It looks like this:

The font used for the letters is called _Ubuntu_.

The icon files for Windows and macOS were generated with the help of https://cloudconvert.com/.
These files are now also used when packaging with electron-packager.


icons/ezra.svg
```diff
@@ -0,0 +1,139 @@
...
+           id="tspan3723"
+           style="font-style:normal;font-variant:normal;font-weight:normal;font-stretch:normal;font-size:7.92806196px;font-family:Ubuntu;-inkscape-font-specification:Ubuntu;font-variant-ligatures:normal;font-variant-caps:normal;font-variant-numeric:normal;font-feature-settings:normal;text-align:start;writing-mode:lr-tb;text-anchor:start;stroke-width:0.3964031px"
+           y="-0.12087331"
+           x="1.8483073"
+           sodipodi:role="line">E</tspan></text>
+      <text
+         id="text3717-0"
+         y="-0.10898132"
+         x="7.4681277"
+         style="font-style:normal;font-variant:normal;font-weight:normal;font-stretch:normal;font-size:7.92806196px;line-height:125%;font-family:'DejaVu Sans';-inkscape-font-specification:'DejaVu Sans';letter-spacing:0px;word-spacing:0px;fill:#000000;fill-opacity:1;stroke:none;stroke-width:0.3964031px;stroke-linecap:butt;stroke-linejoin:miter;stroke-opacity:1"
```
alerque, line 108: This SVG could use some cleanup. I believe Inkscape has a function for this, but I can't remember what it's called. There are a tone of useless specs in here, such as this one that specifies Dejau Sans as the font on an element, only to be unused because the child element specifies Ubuntu. Not only is this a trivial amount of wasted space and a lot of clutter, it also will mean the SVG won't validate for installation without a dependency on DejaVu Sans which it doesn't actually need.

alerque: Actually you know what ... as an icon I thing the right thing to do is actually explode the glyphs into paths. That way the SVG does not bring a new font dependency to the table at all. Inkscape has a facility for this.

tobias-klein: Thanks for the feedback! I just optimized the icon based on your comments:
* Turned the glyphs into paths
* Saved as optimized SVG

Do you have any feedback regarding the design as such? :)

alerque: No offense, but it is kinda ugly. But I'm not a designer and can't do any better, and an icon is better than no icon. I suggest mentioning in the next release announcement that a logo/icon contribution from a designer would be welcome. Combined with the lack-luster nature of this icon that might grab the attention of the kind of person you need to make something better happen. I think having this is place will be better than having nothing and it's far more likely to jog somebody into contributing than having nothing at all.

tobias-klein: Good, that's what I expected :smile: 
I realize that this icon is coming from a programmer, not from a designer ... and I also thought it's better to have something than nothing. To be improved / replaced!

PR #68, opened 2020-06-01
Update Dutch translation

lemtom:


locales/nl/translation.json
```diff
@@ -33,7 +33,8 @@
...
+    "repo-selection-info-text": "Ezra Project werkt met bijbelvertalingsmodules aangeboden door <a class='external' href='http://www.crosswire.org/sword'>het SWORD-project</a> en het <a class='external' href='http://www.crosswire.org'>CrossWire Bijbelgenootschap</a>.<br/><br/>Hieronder ziet u de lijst met SWORD-depots die zijn gepubliceerd door CrossWire. Een depot is een internetopslaglocatie die een set Bijbelvertalingsmodules bevat. Naast elk depot ziet u het totale aantal Bijbelvertalingsmodules dat beschikbaar is uit dat depot. <br/> <br/> Om een bijbelvertalingsmodule te
```
tobias-klein, line 36: There is a line break in the middle of this string that would cause a JSON parsing error.

lemtom: I directly edited from Github sometime. Seems to have been caused by the behavior from the search functionality that it forced upon me (I searched for install and it likely selected it).

PR #69, opened 2020-06-01
Update French translation

lemtom:


lemtom:
Yes. This doesn't seem to be an error I introduced, though the way it happened is possibly very similar.

PR #83, opened 2020-06-12
Find-all-occurrences French translation

lemtom:
I was planning to have a look at it in the weekend, but since it's urgent I've taken a look at it during my coffee break.


tobias-klein:
Thanks, Tom!

PR #153, opened 2020-12-10
Es translations update

reyespinosa1996:
The Spanish translation for 0.16 is waiting for @Rodolinux revision to be merge into Master.


tobias-klein (approved):
Thanks for adding the missing `show-header-navigation` string as well!
I can only judge from a syntax perspective otherwise and leave it to @Rodolinux to approve the translation itself.

Rodolinux (approved):
I approve the changes

PR #182, opened 2021-02-18
Spanish translation updated

reyespinosa1996:
The new strings were translated to Spanish. Maybe our brother @Rodolinux could help us checking if everything is in place.


Rodolinux (approved):
I approve them

Issue #42, opened 2020-04-05
Adjustable font-size for bible text


tobias-klein:

tobias-klein:
The icons could look like this:

![](https://camo.githubusercontent.com/761556584b7476d9e415ba0865dfdd7a44f5b5f1/68747470733a2f2f662e636c6f75642e6769746875622e636f6d2f6173736574732f343332303539332f313134353635312f39623135393938632d316532362d313165332d386233342d6433393265303064313837662e504e47)

zhuiks:
Do you think this could be done directly on menu panel vs inside display option dialog?
If so then it should be a separate component, is that correct?

tobias-klein:
I think it may be good if those plus/minus buttons for the font size would be available in the menu directly. And yes ... it could be a very simple new JS component ... probably just with 2-3 methods (init, enlarge, reduce). Alternatively those methods could simply be integrated in the AppController class.

Would you like to take it on?

tobias-klein:
Oh ... And in any case the font size should be applied globally (for all tabs) and also persisted in the configuration.

PR #197, opened 2021-03-19
Bible text size buttons

zhuiks:
Closing #42, already merged with #195. Please close this only after #195 

That's how it looks:
Screenshot from 2021-03-19 10-17-22

### DIFF
--- a/app/frontend/components/text_size_settings.js
+++ b/app/frontend/components/text_size_settings.js
@@ -1,3 +1,27 @@
+/* This file is part of Ezra Bible App.
+
+   Copyright (C) 2019 - 2021 Tobias Klein <contact@ezra-project.net>
+
+   Ezra Bible App is free software: you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation, either version 2 of the License, or
+   (at your option) any later version.
+
+   Ezra Bible App is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with Ezra Bible App. See the file LICENSE.
+   If not, see <http://www.gnu.org/licenses/>. */
+
+/**
+ * The TextSizeSettings component implements the buttons two control text (font size) of the verse and
+ * other elements. Size update is done by updating (adding new and deleting old) css rule
+ * 
+ * @category Component
+ */
 class TextSizeSettings {
   constructor() {
     this.defaultTextSize = 10; // in em*10 so not to deal with float
@@ -8,7 +32,7 @@ class TextSizeSettings {
     this.openMenuButton = '.text-size-settings';
     this.menuContainer = '.text-size-menu';
     this.menuIsOpened = false;
-    this.stylesheet = null;  // CSSStyleSheet
+    this.stylesheet = null;  // instance of CSSStyleSheet https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet
   }
 
   async init() {
@@ -67,15 +91,14 @@ class TextSizeSettings {
     var buttonOffset = $menuButton.offset();
     var $menuContainer = currentVerseListMenu.find(this.menuContainer);
 
-    var top_offset = buttonOffset.top + $menuButton.height() + 2;
-    var left_offset = buttonOffset.left - 2;
-    if(left_offset+$menuContainer.width() > $(window).width()) {
-      left_offset = buttonOffset.left + $menuButton.width() - $menuContainer.width();
+    var topOffset = buttonOffset.top + $menuButton.height() + 2;
+    var leftOffset = buttonOffset.left - 2;
+    if (leftOffset+$menuContainer.width() > $(window).width()) {
+      leftOffset = buttonOffset.left + $menuButton.width() - $menuContainer.width();
     }
 
-    console.log(`[${buttonOffset}] ->[${top_offset}, ${left_offset}]`);
-    $menuContainer.css('top', top_offset);
-    $menuContainer.css('left', left_offset);
+    $menuContainer.css('top', topOffset);
+    $menuContainer.css('left', leftOffset);
 
     this.menuIsOpened = true;
     $menuContainer.show();
--- a/app/frontend/components/text_size_settings.js
+++ b/app/frontend/components/text_size_settings.js
@@ -16,6 +16,16 @@
    along with Ezra Bible App. See the file LICENSE.
    If not, see <http://www.gnu.org/licenses/>. */
 
+const Mousetrap = require('mousetrap');
+
+const DEFAULT_TEXT_SIZE = 10; // in em*10 so not to deal with float
+const MIN_SIZE = 7;
+const MAX_SIZE = 20;
+const INCREASE_SHORTCUT = ['mod+=', 'mod+shift+=']; // Ctrl/Cmd + 
+const DECREASE_SHORTCUT = ['mod+-', 'mod+shift+-'];  // Ctrl/Cmd -
+const RESET_SHORTCUT = 'mod+0';  // Ctrl/Cmd 0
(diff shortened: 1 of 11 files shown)


tobias-klein:
Can't look at this immediately ... but within the next couple of days!
Thanks for your efforts, I appreciate it!

tobias-klein:
I have just had a first look from a user perspective. It works well - including persistence!

One question that came to my mind though ... wouldn't it make sense to also increase the text-size of the following items simultaneously:
* Verse Reference text (may require also adjusting the overall width of that column, especially for passages like Psalm 119)
* Verse tags
* Verse notes
* Book notes

What do you think?

I'll have a separate look regarding the coding of this.

zhuiks:
> One question that came to my mind though ... wouldn't it make sense to also increase the text-size of the following items simultaneously:
> 
>     * Verse Reference text (may require also adjusting the overall width of that column, especially for passages like Psalm 119)
>     * Verse tags
>     * Verse notes
>     * Book notes
> 
> What do you think?

I think it depends on use cases. I personally prefer to have Bible text bigger as it's easier to concentrate on that. Maybe it can be another display option: "Adjust tags and notes text size along with Bible text".
In any case it's easy to add as current implementation just simply updates css rule.

> Another thing that comes to my mind ... How about adding shortcuts for increasing and decreasing the font-size. Typically it would be CTRL/CMD +/- (Win/Linux vs. Mac). Search the code for `Mousetrap.bind` and you'll find some examples!

Great idea, I'll look into that.

tobias-klein:
> > One question that came to my mind though ... wouldn't it make sense to also increase the text-size of the following items simultaneously:
> > 
> > * Verse Reference text (may require also adjusting the overall width of that column, especially for passages like Psalm 119)
> > * Verse tags
> > * Verse notes
> > * Book notes
> > 
> > What do you think?
> 
> I think it depends on use cases. I personally prefer to have Bible text bigger as it's easier to concentrate on that. Maybe it can be another display option: "Adjust tags and notes text size along with Bible text".
> In any case it's easy to add as current implementation just simply updates css rule.

The Bible text is bigger than the other texts (notes/tags) in general.
However, I guess if somebody uses the app on a really large screen there may be a motivation to adjust all text content in the bible text container. And yes, we could make that configurable in the options menu, good idea!

app/frontend/components/text_size_settings.js
```diff
@@ -1,31 +1,79 @@
...
-    this._settingsObject = window.ipcSettings;
+    this.defaultTextSize = 10; // in em*10 so not to deal with float
+    this.minSize = 7;
+    this.maxSize = 20;
+    this._textSizeValue = this.defaultTextSize;
+    this._settingsKey = 'verse-text-size';
     this.openMenuButton = '.text-size-settings';
     this.menuContainer = '.text-size-menu';
     this.menuIsOpened = false;
+    this.stylesheet = null;  // CSSStyleSheet
```
tobias-klein, line 11: I suggest to add a space in the comment for readability ...
`CSS Stylesheet`

app/frontend/components/text_size_settings.js
```diff
@@ -1,31 +1,79 @@
...
 
-    var top_offset = $menuButton.top + $menuButton.height() + 1;
-    var left_offset = $menuButton.left + $menuButton.width()/2 - $menuContainer.width()/2;    
+    var top_offset = buttonOffset.top + $menuButton.height() + 2;
+    var left_offset = buttonOffset.left - 2;
+    if(left_offset+$menuContainer.width() > $(window).width()) {
+      left_offset = buttonOffset.left + $menuButton.width() - $menuContainer.width();
+    }
+
+    console.log(`[${buttonOffset}] ->[${top_offset}, ${left_offset}]`);
```
tobias-klein, line 76: I suggest to out comment the console.log before merging.

app/frontend/components/text_size_settings.js
```diff
@@ -1,31 +1,79 @@
...
+    var $menuButton = currentVerseListMenu.find(this.openMenuButton);
     $menuButton.addClass('ui-state-active');
+    var buttonOffset = $menuButton.offset();
     var $menuContainer = currentVerseListMenu.find(this.menuContainer);
 
-    var top_offset = $menuButton.top + $menuButton.height() + 1;
-    var left_offset = $menuButton.left + $menuButton.width()/2 - $menuContainer.width()/2;    
+    var top_offset = buttonOffset.top + $menuButton.height() + 2;
+    var left_offset = buttonOffset.left - 2;
+    if(left_offset+$menuContainer.width() > $(window).width()) {
```
tobias-klein, line 72: Code style: Insert a space - `if (`

app/frontend/components/text_size_settings.js
```diff
@@ -1,31 +1,79 @@
...
-    var $menuButton = currentVerseListMenu.find('.display-options-button');
+    var $menuButton = currentVerseListMenu.find(this.openMenuButton);
     $menuButton.addClass('ui-state-active');
+    var buttonOffset = $menuButton.offset();
     var $menuContainer = currentVerseListMenu.find(this.menuContainer);
 
-    var top_offset = $menuButton.top + $menuButton.height() + 1;
-    var left_offset = $menuButton.left + $menuButton.width()/2 - $menuContainer.width()/2;    
+    var top_offset = buttonOffset.top + $menuButton.height() + 2;
+    var left_offset = buttonOffset.left - 2;
```
tobias-klein, line 71: Code style: Let's try to be consistent using camel case for all variables. There is some legacy code where this is not the case, but let's try to use this style for all new code.

app/frontend/components/text_size_settings.js
```diff
@@ -1,31 +1,79 @@
 class TextSizeSettings {
```
tobias-klein, line 1: Please add a small API docs summary above the class definition

tobias-klein (commented):
Thanks for this!

I only have a few comments, mostly style-related. Your code is easy to read and clean! 👍

html/verse_list_tabs.html
```diff
@@ -1,79 +1,119 @@
...
+      <div class="text-size-menu verse-selection-menu" style="display: none;">
+        <div class="verse-list-menu-line">
+          <div class="fg-button text-size-decrease ui-state-default ui-corner-all" i18n="[title]menu.text-size-decrease">
+            <i class="fas fa-font"></i><i class="fas fa-minus fa-xs"></i>
+          </div>
+          <div class="fg-button text-size-reset ui-state-default ui-corner-all" i18n="[title]menu.text-size-reset">
+            <i class="fas fa-font"></i>
+          </div>
+          <div class="fg-button text-size-increase ui-state-default ui-corner-all" i18n="[title]menu.text-size-increase">
+            <i class="fas fa-font"></i><i class="fas fa-plus fa-xs"></i>
```
tobias-klein, line 47: Would you mind adding the new i18n texts to all the locales files?

zhuiks:
I added all those changes we were talking about including keyboard shortcuts and display option.
@tobias-klein what should adjust text size buttons dialog do when user switches to another tab or clicks somewhere else?

tobias-klein:
Thank you! Excellent!
I think the menu should simply be hidden when clicking somewhere else, also when switching to another tab. I believe there is already functionality like this implemented for other menus and there is a central function that hides all menus.

zhuiks:
> I think the menu should simply be hidden when clicking somewhere else, also when switching to another tab. I believe there is already functionality like this implemented for other menus and there is a central function that hides all menus.

This should do it

tobias-klein:
Excellent!

tobias-klein:
It works very well! Excellent job.

I have one more suggestion. We can disable/enable the buttons (add/remove class `ui-state-disabled`) for the following cases:

* Initial state: The 'reset to default' button can be disabled, because it is the default already. Once changing the font-size the button can be enabled, once resetting it disabled again.
* Once reaching the maximum font-size the increase button can be disabled. Once going down a notch it can be enabled again.
* Once reaching the minimum the decrease button can be disabled. Once going up a notch it can be enabled again.

zhuiks:
> I have one more suggestion. We can disable/enable the buttons (add/remove class `ui-state-disabled`)

Great idea! Here you go: 033f0ed

tobias-klein (approved):

PR #212, opened 2021-04-06
Refactor tests

zhuiks:
I have done some refactoring... )))
Basically I just split spectron_helper into multiple helpers. 
I also removed class wrap for helper as module itself is wrapped in a function by node.js
I also tried to minimize global variables.
I hope this all will make code more readable and easier to maintain all the tests


tobias-klein:
Thanks for this! Good job.

PR #223, opened 2021-04-12
Restore scroll position

tobias-klein:
@zhuiks Would you mind testing this a bit?

There are three scenarios:

* Restoring scroll position when switching between existing tabs.
* Restoring scroll position when switching to the previous tab after adding a new tab.
* Restoring scroll position after app start.

### DIFF
--- a/app/frontend/controllers/tab_controller.js
+++ b/app/frontend/controllers/tab_controller.js
@@ -37,6 +37,11 @@ class TabController {
     this.metaTabs = [];
     this.loadingCompleted = false;
     this.lastSelectedTabIndex = null;
+
+    // The scrollTop value returned from the verse list frame element needs some adjustment
+    // when being restored. This value is an offset that will be applied at the time the
+    // scroll position is restored.
+    this.SCROLL_OFFSET = 230;
   }
 
   init(tabsElement, tabsPanelClass, addTabElement, settings, tabHtmlTemplate, onTabSelected, onTabAdded, defaultBibleTranslationId) {
@@ -73,6 +78,9 @@ class TabController {
 
     exitContext.addEventListener(exitEvent, async () => {
       console.log('Persisting data!');
+
+      this.lastSelectedTabIndex = 0;
+      this.savePreviousTabScrollPosition();
 
       await this.saveTabConfiguration();
       await this.saveBookSelectionMenu();
@@ -377,15 +385,7 @@ class TabController {
           
           var index = this.getCorrectedIndex(ui);
 
-          if (this.lastSelectedTabIndex != null) {
-            var previousVerseListFrame = app_controller.getCurrentVerseListFrame(this.lastSelectedTabIndex);
-            var previousTab = this.getTab(this.lastSelectedTabIndex);
-
-            if (previousTab != null) {
-              const SCROLL_OFFSET = 230;
-              previousTab.setScrollTop(previousVerseListFrame[0].scrollTop + SCROLL_OFFSET);
-            }
-          }
+          this.savePreviousTabScrollPosition();
 
           if (metaTab.getTextType() != null) {
             var currentVerseList = app_controller.getCurrentVerseList(index);
@@ -411,8 +411,7 @@ class TabController {
               var currentVerseList = app_controller.getCurrentVerseList(index);
               currentVerseList.show();
 
-              var currentVerseListFrame = app_controller.getCurrentVerseListFrame(index);
-              currentVerseListFrame[0].scrollTop = metaTab.getScrollTop();
+              this.restoreScrollPosition(index);
               app_controller.hideVerseListLoadingIndicator(index);
             }
           }
@@ -435,6 +434,36 @@ class TabController {
     });
 
     uiHelper.configureButtonStyles('.ui-tabs-nav');
+  }
+
+  savePreviousTabScrollPosition() {
+    if (this.lastSelectedTabIndex != null) {
+      var previousVerseListFrame = app_controller.getCurrentVerseListFrame(this.lastSelectedTabIndex);
+      var previousTab = this.getTab(this.lastSelectedTabIndex);
+
+      if (previousTab != null) {
+        previousTab.setScrollTop(previousVerseListFrame[0].scrollTop + this.SCROLL_OFFSET);
+      }
+    }
+  }
+
+  restoreScrollPosition(tabIndex, isStartup=false) {
+    var metaTab = this.getTab(tabIndex);
+
+    if (metaTab != null) {
+      var currentVerseListFrame = app_controller.getCurrentVerseListFrame(tabIndex);
+
+      if (currentVerseListFrame != null) {
+        var savedScrollTop = metaTab.getScrollTop();
+        if (isStartup) {
+          // On startup the scroll offset needs to be removed.
+          // It is unclear why ... but this is how it works!
+          savedScrollTop -= this.SCROLL_OFFSET;
+        }
+
+        currentVerseListFrame[0].scrollTop = savedScrollTop;
+      }
+    }
   }
 
   getSelectedTabIndex() {
--- a/app/frontend/controllers/tab_controller.js
+++ b/app/frontend/controllers/tab_controller.js
@@ -459,7 +459,7 @@ class TabController {
       var currentVerseListFrame = app_controller.getCurrentVerseListFrame(tabIndex);
 
       if (currentVerseListFrame != null) {
-        var savedScrollTop = metaTab.getLocation();
+        const savedScrollTop = metaTab.getLocation();
 
         if (savedScrollTop != null) {
           // console.log("Setting location to " + savedScrollTop);
--- a/app/frontend/controllers/tab_controller.js
+++ b/app/frontend/controllers/tab_controller.js
@@ -384,7 +384,9 @@ class TabController {
           
           var index = this.getCorrectedIndex(ui);
 
-          this.savePreviousTabScrollPosition();
+          if (metaTab.selectCount > 1) {
+            this.savePreviousTabScrollPosition();
+          }
 
           if (metaTab.getTextType() != null) {
             var currentVerseList = app_controller.getCurrentVerseList(index);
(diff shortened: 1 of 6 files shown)


zhuiks:
>  - Restoring scroll position when switching between existing tabs.
>  - Restoring scroll position when switching to the previous tab after adding a new tab.

These are working  just fine. The only thing I noticed it takes sometimes to upload a text when I'm switching back to previous tab. Especially if it's a bigger book. But I think it's from other commits, because this behaviour is already in master

>  - Restoring scroll position after app start.
Having multiple tabs, first tab would not restore. But if it's only one tab it works

tobias-klein:
> > * Restoring scroll position when switching between existing tabs.
> > * Restoring scroll position when switching to the previous tab after adding a new tab.
> 
> These are working just fine. The only thing I noticed it takes sometimes to upload a text when I'm switching back to previous tab. Especially if it's a bigger book. But I think it's from other commits, because this behaviour is already in master

Switching to bigger books already took some time before. The difference is that now you the tab switching is more fluent and you get a loading indicator. There is actually no "loading" happening, but it's just the hide/show of a large portion of the DOM (which is slow!).

> 
> > * Restoring scroll position after app start.
> 
> Having multiple tabs, first tab would not restore. But if it's only one tab it works

Ok, good that you found that. I'll look into that.

tobias-klein:
@zhuiks Could you check once more? The last commit should fix the issue you found.

zhuiks:
Now it works!

But I found inconsistency when resizing the window:
- I have several tabs open
- When I resize a window and switch to other tab:
The position does shift. The farther down I'm in the book the bigger shift is. Then if I switch to the first tab (that was open on resize) position also shifts, although it has was fine right after resizing. 

That's noticeable when I quit the app and open it again and app has a different window size

tobias-klein:
Currently this is handled based on a absolute scroll position (coming from the scrollTop attribute). In order to make this independent from the window size, I need to find a way to calculate a relative scroll position. I'll look into this!

zhuiks:
> Currently this is handled based on an absolute scroll position (coming from the scrollTop attribute). In order to make this independent from the window size, I need to find a way to calculate a relative scroll position. I'll look into this!

Maybe it's worth to save the top visible verse-referene-id of the verse box and calculate scrollTop from there. In that case in the future we can request verses relative to that verse for faster visual feedback

tobias-klein:
> > Currently this is handled based on an absolute scroll position (coming from the scrollTop attribute). In order to make this independent from the window size, I need to find a way to calculate a relative scroll position. I'll look into this!
> 
> Maybe it's worth to save the top visible verse-referene-id of the verse box and calculate scrollTop from there. In that case in the future we can request verses relative to that verse for faster visual feedback

The thing is that an absolute position of scrollTop in pixels is only valid for a given window size.
What we could do is use the [Document.elementFromPoint](https://developer.mozilla.org/en-US/docs/Web/API/Document/elementFromPoint) method to calculate the top-most element visible in the viewport. I don't know if this will work, have never done this before. I'll have another look in the next days.

tobias-klein:
@zhuiks 
I have changed the approach for saving the location. This is now done based on the first visible verse in the verse list (see helper method UiHelper.getFirstVisibleVerseAnchor. Could you test once more?

This is not as precise as scrollTop, but still precise enough I would say. The scoll position may change slightly. But the window size changes do not matter any longer.

The new helper method could also be used for other use cases, like updating the navigation bar with the current chapter more precisely than it is done right now (currently it is based on verse-box mousover).

app/frontend/helpers/ui_helper.js
```diff
@@ -216,6 +216,66 @@ class UiHelper {
...
+
+          currentElement = currentElement.previousElementSibling;
+        }
+      } else {
+        // We are dealing with an element inside a verse-box
+        const MAX_ELEMENT_NEXTING = 7;
+
+        // Traverse up the DOM to find the verse-box
+        for (var i = 0; i < MAX_ELEMENT_NEXTING; i++) {
+          if (currentElement.classList != null && currentElement.classList.contains('verse-box')) {
```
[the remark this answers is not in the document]
zhuiks: If I close inactive tab the error doesn't occur

app/frontend/helpers/ui_helper.js
```diff
@@ -216,6 +216,66 @@ class UiHelper {
...
+    if (verseListFrame != null) {
+      var verseListFrameRect = verseListFrame[0].getBoundingClientRect();
+
+      var currentNavigationPane = app_controller.navigation_pane.getCurrentNavigationPane()[0];
+      var currentNavigationPaneWidth = currentNavigationPane.offsetWidth;
+
+      var firstElementOffsetX = verseListFrameRect.x + currentNavigationPaneWidth + 15;
+      var firstElementOffsetY = verseListFrameRect.y + 15;
+      
+      var firstElement = document.elementFromPoint(firstElementOffsetX, firstElementOffsetY);
```
zhuiks, line 233: Maybe declare these variables with const/let as they are only relevant to the current block scope

app/frontend/helpers/ui_helper.js
```diff
@@ -216,6 +216,66 @@ class UiHelper {
...
+      var currentElement = firstElement;
```
zhuiks, line 234: I propose to declare `var currentElement` on top of the function. That way it would be understandable that it is "global" for the function scope (actually all `var`s are visible in the whole function scope. Doesn't matter where they are declared).

app/frontend/helpers/ui_helper.js
```diff
@@ -216,6 +216,66 @@ class UiHelper {
...
+            currentElement.previousElementSibling.nodeName == 'A') {
+
+          currentElement = currentElement.previousElementSibling;
+        }
+      } else {
+        // We are dealing with an element inside a verse-box
+        const MAX_ELEMENT_NEXTING = 7;
+
+        // Traverse up the DOM to find the verse-box
+        for (var i = 0; i < MAX_ELEMENT_NEXTING; i++) {
```
zhuiks, line 256: Maybe use `let i = 0;` for it to be in more "local" block scope

app/frontend/helpers/ui_helper.js
```diff
@@ -216,6 +216,66 @@ class UiHelper {
...
+    var verseListFrame = app_controller.getCurrentVerseListFrame();
+
+    if (verseListFrame != null) {
+      var verseListFrameRect = verseListFrame[0].getBoundingClientRect();
+
+      var currentNavigationPane = app_controller.navigation_pane.getCurrentNavigationPane()[0];
+      var currentNavigationPaneWidth = currentNavigationPane.offsetWidth;
+
+      var firstElementOffsetX = verseListFrameRect.x + currentNavigationPaneWidth + 15;
+      var firstElementOffsetY = verseListFrameRect.y + 15;
```
zhuiks, line 231: Can use please add a comment for the "magic number". Maybe these numbers are more relevant in the next line, i.e.
```
const firstElement = document.elementFromPoint(firstElementOffsetX + 15, firstElementOffsetY + 15);
```

app/frontend/controllers/tab_controller.js
```diff
@@ -422,6 +437,38 @@ class TabController {
...
+    if (this.lastSelectedTabIndex != null) {
+      this.saveTabScrollPosition(this.lastSelectedTabIndex);
+    }
+  }
+
+  restoreScrollPosition(tabIndex) {
+    var metaTab = this.getTab(tabIndex);
+
+    if (metaTab != null) {
+      var currentVerseListFrame = app_controller.getCurrentVerseListFrame(tabIndex);
```
zhuiks, line 459: Maybe `const currentVerseListFrame` would be better

app/frontend/controllers/tab_controller.js
```diff
@@ -422,6 +437,38 @@ class TabController {
...
+
+      if (currentVerseListFrame != null) {
+        var savedScrollTop = metaTab.getLocation();
```
zhuiks, line 462: Maybe `const savedScrollTop` would be better

zhuiks (commented):
I'm getting one error.

And I have an opinion that using var inside if/for blocks is misguiding because it declares variable in the function scope

zhuiks:
Other than one error, everything works great. It saves position even with resizing! :+1:

app/frontend/helpers/ui_helper.js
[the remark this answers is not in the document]
tobias-klein: Ok, thanks for reporting this!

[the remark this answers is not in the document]
tobias-klein: This is now fixed in the latest commit

app/frontend/helpers/ui_helper.js
tobias-klein: currentElement is only used in the inner block of the function (within the condition `if (verseListFrame != null && verseListFrame.length > 0) {`). I usually don't put all "block-scope" variables on top.

app/frontend/helpers/ui_helper.js
tobias-klein: Good point.

app/frontend/helpers/ui_helper.js
tobias-klein: Good point. I guess I should consider this for the whole code base ... you'll find this pattern with `for (var i = 0; ...` all over the place.

tobias-klein: This is now fixed in the latest commit.

app/frontend/helpers/ui_helper.js
tobias-klein: Done

app/frontend/controllers/tab_controller.js
tobias-klein: `const` on Objects or Arrays actually does not prevent modifying them. It merely fixes the variable to the object it is pointing to. With `const` I actually see the point more for numeric values or strings, not so much for Objects and Arrays.

app/frontend/controllers/tab_controller.js
tobias-klein: Yes!

Issue #214, opened 2021-04-07
New app logo


zhuiks:
Maybe the logo should be a little bit more interesting.
How about something like this:

------
Update: removed first attempt to the comment below so not to spoil the first impressions

Update2: The latest version:
![logo8](https://user-images.githubusercontent.com/3584119/114562857-80ea4000-9c34-11eb-938f-902da1e6ca09.png)

zhuiks:
@alerque What do you think?

alerque:
Let me get out my microscope...

I'd say this is an improvement that leaves room for improvement. I'm not much of a designer type myself so take my comments with the proverbial grain. Perhaps the thing that bothered me most was how long it took me to figure out what the blobs were in front of the scroll. Obvious once you see it, but it took me a while. Maybe it was that the head is disembodied, maybe the person could use some treatment such as maybe an outline with a lighter fill.

Either way if your focus is a little blurry it's hard for the scroll not to look like the dude 's got his hands in the air.

Sorry if that's too critical. I actually like it, but at the same time it could use some tweaking. But I'm not qualified to do the tweaking, so perhaps seek the opinion of somebody with the right skill set.

zhuiks:
Thank you for your critical review! In my case it's inspiring :smile: 

> Let me get out my microscope...

Hmm, I didn't realize it would be so small on the phones...

Here is another more minimalistic approach: 
![Logo 3](https://user-images.githubusercontent.com/3584119/114056589-70fae680-9857-11eb-84ce-0ef5b8894e3a.png) 
![Logo 3(1)](https://user-images.githubusercontent.com/3584119/114057650-59702d80-9858-11eb-97c2-d98fc4f1818f.png)
![Logo 3](https://user-images.githubusercontent.com/3584119/114057623-537a4c80-9858-11eb-9a62-e36c99584f81.png)

alerque:
I wasn't an a phone, I was on a massive desktop monitor spread ;-)

First impressions of the more minimalist approach: Fantastic! Of course I've lost the benefit of truly first impressions, this time I was expecting a person reading a scroll, but the simplicity still gets the idea across with getting quite so muddled in the process.

For larger sizes (and resolutions) where we get more to work with (such as mobile app icons) perhaps more detail would be in order, but I like the way this scaled down.

The scrolls at the side could perhaps use something solid in them, at least going down if not protruding anywhere. They look a bit like waves...

tobias-klein:
[...]

Issue #171, opened 2021-01-16
Change of product name and transfer of domain - part 1


tobias-klein:
Another Ezra Project requested the change of the product name and along with that also the transfer of the domain ezra-project.net. The justification given is longer existence of the other project.

I expressed openness for this kind of change.
I already have a new name on my mind and will work on this change soon.

- [x] Request new domain names
- [x] Change name within product source code
- [x] Adjust GitHub CI build configuration
- [x] Create script for migrating database from previous folder to new folder
- [x] Create script for migrating configuration from previous folder to new folder
- [x] Change app id for Android app
- [x] Update URL in apidocs
- [x] Update URL in NewReleaseChecker
- [x] Change name in README
- [x] Change GitHub organization name
- [x] Change all repo names
- [x] Change name and link in CrossWire Wiki
- [x] Change name and link on Wikipedia pages
- [x] Change name and logo in CrossWire News
- [x] Rename distro packages in Arch Linux

PR #225, opened 2021-04-13
New logo

zhuiks:
Logo files from #214 discussion
- png and ico are 256x256
- icns is converted from 512x512

Let me know if we need other sizes

Maybe it can be a part of changing product name #171


tobias-klein:
I like the logo. It doesn't look so nice in the scaled down version on Windows, though. Windows is the primary user group. Do you see a possibility to optimize a bit more?


image

alerque:
I see what you mean about looking bad in that Windows screen shot, but ironically the screenshot in the issue scaled to the width of the GitHub issues column looks great. In other words it's not just the size that's a problem as it scales much smaller just fine, it's something specifically unlucky about the rasterization size and settings being used and it can surely be fixed ... by hand hinting if we have to but by design and export tweaks too.

alerque:
I suggest also having some sort of build rules around so that the icons can be regenerated from the SVG. This tooling wouldn't need to be part of the every day build procedure since the raster resources are also in the repository but at least a reproducible chain of rules to get the output again if something were to change would be handy.

zhuiks:
I can tweak the width of the scroll outline. I notice that a bit thicker width looks better on the small sizes. 

But I agree that the main problem is auto scaling. I just figured out that different icon sizes (16x16, 32x32, 48x48, and 256x256) can be (should be) combined to one ico file. In the previous commit I just converted 256x256 png to ico. So I'll try combining

zhuiks:
@tobias-klein how about now?
I used imagemagick `convert`: 
```
convert icons/ezra.png -define icon:auto-resize="256,48,32,16" icons/ezra.ico
```
If we use it in build script for Windows in github actions does it need to be installed?

How about MacOS how new logo is looking there?

tobias-klein:
Thanks for your hard work on this! I'll review the updates tomorrow!

alerque:
We should be able get Image Magick running in the CI builders just fine. I would note that the `convert` interface from v6 and prior is deprecated in favor of `magick` in v7+. I suggest something in `build_scripts` to regenerate all the resources that are derived from the SVG. It could even be setup so each platform only generated the relevant resources, although it will take a little tweaking to get the right workflow on all systems. At the very least having it scripted in a way developers can refresh the resources will make future tweaks easier. Whether or not the resulting PNGs should even be tracked in the repo is up for debate, for now I say leave them until such a time as they are built per-platform.

tobias-klein:
This is how it looks on macOS:

**DMG Installer**
Bildschirmfoto 2021-04-14 um 19 15 51

**Dock icon**
Bildschirmfoto 2021-04-14 um 19 18 48

tobias-klein:
** Update on Windows**
image

image

tobias-klein:
Does the Mac version intentionally have a black background?

zhuiks:
> Does the Mac version intentionally have a black background?

No it should be transparent, I converted it with [cloud converter](https://cloudconvert.com/png-converter)
Let me try again

zhuiks:
according to [this](https://www.codeproject.com/articles/1067619/what-icon-sizes-are-used-by-windows) standard sizes for Windows are 16, 32, 48, 256 (must have) and additional are 20, 24, 40, 60, 72
```bash
convert icons/windows-source/*.png icons/ezra.ico
```

according to [this](https://gist.github.com/zlbruce/883605a635df8d5964bab11ed75e46ad) and playing with `png2icns` for MacOS sizes are 16, 32, 128, 256, 512, 1024
```bash
sudo apt install icnsutils
png2icns icons/ezra.icns icons/mac-os-source/*.png
```

zhuiks:
I don't have any access to Windows or MacOS computer now. You have to test it. Sorry, guys

tobias-klein:
> I don't have any access to Windows or MacOS computer now. You have to test it. Sorry, guys

Something for the weekend! :) No worries.

tobias-klein:
**Android** 
image

image

It seems like the lower edge could need some anti aliasing.
On Android the PNG version is used.

zhuiks:
> It seems like the lower edge could need some anti aliasing.
On Android the PNG version is used.

Yes, I'm not satisfied. Is it from the latest commit? How do you build for Android?

zhuiks:
I got a hold of the Windows, result is still similar in the for the window header. Although app icon of the same size looks nice:

ezra-windows-screenshot

tobias-klein:
The icon in the top left corner of the window was already looking mediocre before ... (with the old logo). We may need to accept that for the time being. Maybe other open source Electron apps have solved this somehow, could be interesting to have a look at some other apps. I don't think it's a blocking issue for merging this pull request. Let's iterate!
Are you still looking into any other enhancements regarding the icon? Otherwise I would consider to merge this and have it released in **0.18.0**.

tobias-klein:
The icon you saw on Android was from the latest icons/ezra.png.

zhuiks:
> Are you still looking into any other enhancements regarding the icon? Otherwise I would consider to merge this and have it released in 0.18.0.

I'll research a bit about that little corner icon. And I'll try to make logo nicer on Android. 

Let me know how the latest logo looks on MacOS

tobias-klein:
> Let me know how the latest logo looks on MacOS

Bildschirmfoto 2021-04-17 um 08 43 23

Pretty good I would say! Note that I still need to change the name to Ezra Bible App here.

tobias-klein:
Latest update on Android:

image

I feel like the inner rectangle is not necessary. I'd prefer to have the icon a bit bigger. What do you think?

zhuiks:
What version of Cordova are we using?

Which version of Android/launcher are you testing with?

Seems like icons for Android are depending on the launcher.

tobias-klein:
Cordova 7
Default Android Launcher

I am testing on two devices, one running Android 10, the other Android 7.
Why not a newer Cordova version? Because this is what nodejs-mobile requires. See https://code.janeasystems.com/nodejs-mobile/getting-started-cordova

zhuiks:
I followed this in `cofig.xml`:
https://cordova.apache.org/docs/en/7.x/config_ref/images.html#android

Newer versions of Android support adaptive icons with two foreground/background layers. Which was also generated by default with Android studio's asset manager.

But I think we are good with the round one.

tobias-klein:
The adaptive icons are not yet supported by Cordova 7. It seems like Cordova 9 is required for that (see https://github.com/ionic-team/cordova-res/issues/34). Anyway ... I just checked once more on Android and the issue about the missing icon in the app switcher was simply a caching problem. When I cleanly regenerated the app with the latest sources everything looked fine (using the round icon). Therefore I think we are ready to merge now! Thanks for your hard work on this @zhuiks! This is awesome :+1:

tobias-klein (approved):

PR #231, opened 2021-04-17
Translation to Spanish for the 0.18 version.

reyespinosa1996:
I leave unaltered the Ezra Bible App brand on every string because it will be funny to translate it into "Aplicación Bíblica Esdras".
Git commit string was also unaltered because Spanish doesn't have a proper way to translate it and keep the original meaning.
@Rodolinux could you please check if everything is well translated and correct possibles mistakes?


Rodolinux (approved):
I agree with Reinaldo to leave the name of the app in English. In Spanish it sounds a bit ridiculous.

PR #242, opened 2021-05-01
Enhanced in-tab search layout

tobias-klein:
Summary of changes:

- The search form has been moved below the Bible browser box.
- An option has been added to show the search form permanently (useful on Android).
- If the search form is not shown permanently it can still be enabled with CTRL/CMD + f and hidden with Escape.
- The input now stays focussed when pressing ENTER or SHIFT + ENTER.
- The input is now highlighted while focussed.
- Some general bug fixes in the tab search algorithm related to bugs with Strong's modules.


tobias-klein:
@zhuiks Could you test this, please? :)

zhuiks:
> @zhuiks Could you test this, please? :)

Yes, I'll be able to do it on Monday

zhuiks:
Oh, it's a nice feature!
Знімок екрану з 2021-05-03 10-58-54

Everything is working except when I check case-sensitive box or select few words/exact phrase it doesn't clear the previous results. 

When I'm searching with a few words option result count seems to count each word:
Знімок екрану з 2021-05-03 11-14-02

app/frontend/components/tab_search/tab_search.js
```diff
@@ -308,9 +343,14 @@ class TabSearch {
     
     var searchHl = $(verseElements).find('.search-hl, .current-hl');
     for (var i = 0; i < searchHl.length; i++) {
-      var highlightedText = $(searchHl[i]);
-      var text = highlightedText.text();
-      highlightedText.replaceWith(text);
+      var highlightedText = $(searchHl[i])[0];
+      var text = document.createTextNode(highlightedText.innerText);
```
zhuiks, line 347: Let's do let/const inside the block scope. The same is with `for (let i...`

tobias-klein:
Thanks for testing!

> Everything is working except when I check case-sensitive box or select few words/exact phrase it doesn't clear the previous results.

Ok, I'll look into that.
 
> When I'm searching with a few words option result count seems to count each word:

Yes, that's how it is designed at the moment. I did the implementation a while back and cannot remember why, but I am sure I had a reason.

app/frontend/components/tab_search/tab_search.js
tobias-klein: I made some adjustments just now!

locales/en/translation.json
```diff
@@ -200,6 +200,7 @@
     "show-dictionary": "Show dictionary",
     "show-book-chapter-navigation": "Show book/chapter navigation",
     "show-header-navigation": "Show header navigation",
+    "show-tab-search": "Show tab search",
```
zhuiks, line 203: Maybe "Show **current** tab search" is more descriptive.

zhuiks (approved):
Except the string for the display option everything looks good.

zhuiks:
From UX standpoint the little Search icon in the bottom left or top right corner which would convert into the current search overlay on click/touch would be the best. 
But maybe we can save it for future refactoring as the web component.

locales/en/translation.json
tobias-klein: Let's ask a native speaker ... @alerque What do you think?

This is for an option that toggles the search form which allows the user to search within the currently active tab.

alerque: The expression "show tab search" suggests to me a quest to find that tiny tab you left open in a sea of tabs. I had to check several times to confirm you meant something else because my brain didn't want to break out of that assumption. Given your description "Show current tab search" would be better, or maybe "Show search in current tab" as a more natural reading.

That being said I may be a "native" speaker but having lived in Turkey for closing in on two decades and being married to a native Russian speaker hasn't done wonders for my English. Much public shade has been cast on my native speaker credentials.

PR #247, opened 2021-05-03
Config option component

tobias-klein:
This turns the existing DisplayOption class into a real web component (now called ConfigOption).

Quite some boilerplate code could be removed from the display_options_menu.html.

In the future exact rendering and style of the ConfigOption can be easily adjusted based on the existing web component.


app/frontend/components/config_option.js
```diff
@@ -0,0 +1,122 @@
...
+
+class ConfigOption extends HTMLElement {
+  constructor() {
+    super();
+
+    this.innerHTML = `
+      <div class="switch-box fg-button fg-button-toggleable ui-corner-all ui-state-default">
+        <input class="toggle-config-option-switch" type="checkbox"></input>
+        <div class="switch-label"></div>
+      </div
```
zhuiks, line 27: missing '>'

zhuiks:
Wow, seems like code became more clear. And it seems like interface became faster! :clap: 

css styles for `.switch-box` and `.switch-label` should also go into the component.
After that I guess goes refactor of `options_menu.js`
I'll post some additional thoughts under discussion

tobias-klein:
The only CSS part that currently remains in main.css is this, a style that is shared between .tag-select-label, .module-search-label and .switch-label - I would like to keep .switch-label here for now:

```css
.switch-label,
.tag-select-label,
.module-search-label {
  margin: 0.2em;
  margin-left: 0.2em;
}
```

@zhuiks Any other comments before I merge? (Please approve)

zhuiks (dismissed):
Everything looks great.
As we learn more about web components we might do some refactoring. But that's a great starting point

PR #266, opened 2021-05-07
Refactor utility functions

zhuiks:
I moved few functions (html, waitUntilIdle) from global `window` props to separate module.
That way it's much easier to track what's going on in the code. 
I don't think there any bundle size or performance implications

I think I refactored all the files that mentioned those functions


tobias-klein (dismissed):

tobias-klein:
Thank you. I was thinking of doing this at some point and this refactoring step totally makes sense :)

PR #313, opened 2021-05-24
i18nController / Cordova fixes

tobias-klein:
@zhuiks 

The latest i18nController does not work on Cordova because of the early call to ipcSettings ... 
(not yet available when i18n is initialized on Cordova).

I will check whether I can found a solution.


tobias-klein:
This has been directly fixed on master.

Issue #198, opened 2021-03-19
Add integrated online help that gives an overview for all keyboard shortcuts


tobias-klein:
The information could be integrated in a separate tab of the info dialogue, or in a separate dialogue via a separate button.

Since the shortcuts are platform-specific (at least the CTRL vs CMD key, the content needs to a bit dynamic based on the platform. CTRL is used on Windows and Linux, while CMD is used on macOS.

The following shortcuts need to be documented:

| Summary                                        | Shortcut                                                                           |
| ------------------------------------------- | ---------------------------------------------------------------------- |
| In tab search                                  | CTRL/CMD + f                                                                 |
| Hide in tab search form                  | ESC                                                                                 |
| Jump to next search occurance     | Enter                                                                                |
| Jump to previous occurance          | Shift + Enter                                                                     |
| Copy selected verses to clipboard | CTRL/CMD + c                                                                 |
| Toggle fullscreen (only Win/Linux) | F11                                                                                   |
| Enable dynamic Strong's display   | Shift + mouse over verse while dictionary info is opened |
| Open new tab                                | CTRL/CMD + t                                                                  |

The texts need to be localized.

Short cut references of other software can serve as an inspiration.

zhuiks:
Shortcuts for #42 Adjustable font size:
| Summary | Shortcut |
| ------------- | ----------- |
| Increase current font size | `CTRL/CMD +` |
| Decrease current font size | `CTRL/CMD -` |
| Reset font size to default | `CTRL/CMD 0` |

augustin-colesnic:
**I picked this up. WIP**
Pushing first changes as in the screenshot attached:

<img width="832" alt="Keboard Shortcuts tab" src="https://user-images.githubusercontent.com/592416/119560248-5e5b5300-bdac-11eb-96fa-92306f21ab6b.png">

augustin-colesnic:
Aaaaahhh! I forgot to create a feature branch and i committed on my forked master. 
Is it ok to create a PR from my forked master?

augustin-colesnic:
pls Review MR https://github.com/ezra-bible-app/ezra-bible-app/pull/316

tobias-klein:
Nice! I'll give feedback in the PR!

PR #316, opened 2021-05-25
Add integrated online help that gives an overview for all keyboard shortcuts

augustin-colesnic:
Covering #198


tobias-klein:
Thank you for this contribution!

One general feedback: What's the idea behind putting the actual shortcuts into the locale? The shortcuts themselves are not different per language, are they? How about putting in the special keys like shift, ctrl, etc. but not the full short cut? Then we can assemble the translations dynamically in the template.

Visually I would suggest to align the column header text left.

augustin-colesnic:
> One general feedback: What's the idea behind putting the actual shortcuts into the locale? The shortcuts themselves are not different per language, are they? How about putting in the special keys like shift, ctrl, etc. but not the full short cut? Then we can assemble the translations dynamically in the template. That would also allow us to add additional shortcuts in the future without having to translate them each time.


**I was wondering the same until I noticed there is on shortcut with explanation on it** (more than the shortcut itself) that states: Shift + mouse over verse while dictionary info is opened. If we get rid of that we could also remove the shortcut section from translation indeed. :)

augustin-colesnic:
So, @tobias-klein what about this long description in the shortcutcolumn for the SHIFT + mouse over? 
Here are the latest changes:
keboward shortcuts Ezra Bible App - 27 may 2021

augustin-colesnic:
Romanian translation for keyboard shortcuts  

romanian translation for keyboard shortcuts

tobias-klein:
The one for "shift plus mouse over" is the only exception, right? So in that case I don't have a problem with a specific entry that needs to be localized, but in all the other cases we can build the shortcut from two independently localized parts, right?

tobias-klein:
Here is another feedback on the visual presentation ...

I think the content is quite easy to read and we do not necessarily need the column headers. However, we could use the left column to insert a "header" whenever a new group of functionality starts ... maybe currently:
1. Font size
2. Current tab search
3. Miscellaneous

There will likely be more shortcuts in the future, so I think we should put them in functionality groups.

tobias-klein:
@augustin-colesnic Any additional feedback on this?

augustin-colesnic:
> Here is another feedback on the visual presentation ...
> 
> I think the content is quite easy to read and we do not necessarily need the column headers. However, we could use the left column to insert a "header" whenever a new group of functionality starts ... maybe currently:
> 
> 1. Font size
> 2. Current tab search
> 3. Miscellaneous
> 
> There will likely be more shortcuts in the future, so I think we should put them in functionality groups.

Oh yes @tobias-klein  those are good observations and I will implement them right away. Sorry for the late replay. I was off the radar lately :smile:

augustin-colesnic:
> The one for "shift plus mouse over" is the only exception, right? So in that case I don't have a problem with a specific entry that needs to be localized, but in all the other cases we can build the shortcut from two independently localized parts, right?

Right now there are 2 independently localised parts (column1 and column2). Do you want me to have only the first column translated and leave the second one without the possibility to be localised?

augustin-colesnic:
Latest changes:

Screenshot 2021-06-11 at 20 55 10

tobias-klein:
I'll look in detail later during the weekend! On the road today.

How about some padding between the shortcut groups? At least 0.5 or 1em?

tobias-klein:
> Right now there are 2 independently localised parts (column1 and column2). Do you want me to have only the first column translated and leave the second one without the possibility to be localised?

**Regarding the second column:**

What I intend here is that we translate the actual keyboard keys (CMD / CTRL / SHIFT) ... because there are translations (CTRL for example is STRG in German). But then the letter/character after that does not need a translation, because it is fixed. So with that said, you can have the "control keys" in the locale, and combine them with a fixed letter (not in locale) inside the code. Does that make sense?

This way we can keep the locale a bit leaner and minimize update effort in the future.

augustin-colesnic:
> > Right now there are 2 independently localised parts (column1 and column2). Do you want me to have only the first column translated and leave the second one without the possibility to be localised?
> 
> **Regarding the second column:**
> 
> What I intend here is that we translate the actual keyboard keys (CMD / CTRL / SHIFT) ... because there are translations (CTRL for example is STRG in German). But then the letter/character after that does not need a translation, because it is fixed. So with that said, you can have the "control keys" in the locale, and combine them with a fixed letter (not in locale) inside the code. Does that make sense?
> 
> This way we can keep the locale a bit leaner and minimize update effort in the future.

Nope it does not make sense for me! :flushed: I would say let us leave it locale also for the second column (shortcut one) and if needed to translate (like german) it will be or if it is not needed it will not be. Simpler that way!

augustin-colesnic:
Added the padding (top and bottom) to the section titles

Screenshot 2021-06-12 at 14 24 18

tobias-klein:
> > > Right now there are 2 independently localised parts (column1 and column2). Do you want me to have only the first column translated and leave the second one without the possibility to be localised?
> > 
> > 
> > **Regarding the second column:**
> > What I intend here is that we translate the actual keyboard keys (CMD / CTRL / SHIFT) ... because there are translations (CTRL for example is STRG in German). But then the letter/character after that does not need a translation, because it is fixed. So with that said, you can have the "control keys" in the locale, and combine them with a fixed letter (not in locale) inside the code. Does that make sense?
> > This way we can keep the locale a bit leaner and minimize update effort in the future.
> 
> Nope it does not make sense for me! 😳 I would say let us leave it locale also for the second column (shortcut one) and if needed to translate (like german) it will be or if it is not needed it will not be. Simpler that way!

OK. 👍

tobias-klein:
The only thing I still need to test is how it looks in portrait mode on a 7" tablet. I think you increased the width of the dialog a bit. We may need to make it responsive to screen size. You can also test this my simply resizing the window to a width of 600px or so.

tobias-klein:
Did you add the new keys/texts to all existing locale files? If not, please do.

augustin-colesnic:
> Did you add the new keys/texts to all existing locale files? If not, please do.

No I did not add those to other locale except (EN) and (RO).  
I see the other locale defaults to (EN), please see below screenshot.
1. Shouldn't each translator add those to its own locale? 
2. If you want me to add them,  shall i add the english one to other locale?

Note: I need to increase the with of the window so the tabs should fit in for other locale like French and Spanish:

Screenshot 2021-06-12 at 15 30 38

augustin-colesnic:
Done. @tobias-klein  please review me last 2 commits

tobias-klein:
Yes, we add the English ones to all other locales. This is then becoming a todo for translators. In the future we should add a script that syncs missing entries to all other languages based on the English one. Whenever I ask translators I give them the exact keys. This way it's most convenient for them I think.

augustin-colesnic:
@tobias-klein  what's pending on this one? Can we merge this?

tobias-klein:
Yes. Sorry, I was simply busy with other things on the weekend. The only thing I need to check is whether this is responsive on different tablet screen sizes.

tobias-klein:
@augustin-colesnic Thanks once more! I did a few more optimizations and just merged!

Issue #272, opened 2021-05-08
Use localized language names in module install/uninstall assistant


tobias-klein:

zhuiks:
This overlaps with #230 
Maybe we need to close previous one or merge it together

Issue #238, opened 2021-04-27
Module Install Assistant: put language selection as a first step


zhuiks:
As a user I can read the Bible only in certain languages. So it's much more logical to select which languages I want Bible modules in as a first step in module installation. The second step can show me which repositories have selected languages.

PR #317, opened 2021-05-27
InstallModuleAssistant refactoring

### LINKED ISSUE #201: Module Install Assistant: default/favorite languages on top
It would be nice to have default system language/app locale to display above all other languages, as well as languages for already installed modules

Screenshot from 2021-03-23 10-51-18

zhuiks:
The main idea to make new modules' installation process simpler for a user that is not familiar with The SWORD project.
This should close #201 #238 #272 

TODO:

- [x] Make language selection as a first step
- [x] Refactor steps in separate files/components
- [x] Localize language list if it's possible with vanilla JS
- [x] Language list UI: split languages by categories (Installed, Bible, Most spoken)
- [x] Show installed module languages under 'Used languages' section
- [x] Add locale strings
- [x] Fix language sorting
- [x] Update step intros locale strings
- [x] Refactor all step checkboxes to use new `<assistant-checkbox>` component
- [x] Fix repositories update
- [x] Refactor lock dialog
- [x] Fix/refactor module information
- [x] Check/fix locked module installation
- [x] Check/fix module install cancellation
- [x] Refactor addModuleAssistant to component
- [x] Fix integration test
- [x] Language list UI: add quick search/filter
- [x] Refactor RemoveModuleAssistant
- [x] Sort languages alphabetically in module list step for add and remove
- [x] Add unit test to Github Actions (implemented under #345)
- [x] Always center ModuleAssistant dialog
- [x] Fail properly when there is internet connection problems #390 
- [x] Sync locale updates for all locale files
- [x] selected languages and selected repositories will reset to previously saved state after repository update
- [x] external link is not displayed for locked modules
- [x] performance issue on languages step
- [x] Remove/disable debug logs
- [x] Test on Android
- [x] Do German translation update with new structure to see how it works during translation


zhuiks:
I have something working now.
@tobias-klein can you take a look on the language categories and say your opinion
@augustin-colesnic feel free to check it out also

I still need to do more refactoring

tobias-klein:
Thanks for this! I like the grouping in the language selection!

Here are some usage observations (not giving code feedback yet).

* Used and System Languages: It only shows English for me, even though my installed modules are English, German, Ancient Greek, Spanish
* The repo selection page should show which languages were selected on the previous page.
* The repo checkbox labels on the repo selection page are not clickable any longer (When clicking them the checkbox should be selected - as on the language selection and bible translation selection page).
* The bible translation checkbox labels on the bible translation page are not clickable any longer. 
* When moving to the Bible translations page, the module detail view on the right side initially shows a loading indicator (it should not).
* Loading the Add module assistant the second time generates various JS errors.

Language groups:
* Most Speaking Languages => Should be "Languages with more than 1 module"
* Rear Languages (what does that mean??)

zhuiks:
Thanks for the feedback!
If you look in the code, as for now only [`step_languages.js`](https://github.com/ezra-project/ezra-project/blob/6cfbc364120112e1094d351bb7287be93ee6937c/app/frontend/components/module_assistant/step_languages.js) and [`assistant_checkbox.js`](https://github.com/ezra-project/ezra-project/blob/6cfbc364120112e1094d351bb7287be93ee6937c/app/frontend/components/module_assistant/assistant_checkbox.js) are implemented completely.

### Language Groups:
My idea:

- **"Used and System Languages"** - selected app locale, default system language (if different), all installed module languages
- **"Bible Languages"** - static list: Ancient Greek, Ancient Hebrew, should Hebrew be here?
- **"Most Spoken Languages"** - static list, top ten from [Wikipedia/Ethnologue](https://en.wikipedia.org/wiki/List_of_languages_by_total_number_of_speakers)
- **"Historical Languages"** - languages that have 'ancient', 'extinct', 'historical' type in [ISO-639.3](https://en.wikipedia.org/wiki/ISO_639:a)
- "known-languages" - localized languages with standard Internationalization API (should be all actively used languages), I put it as "Other Languages",
- "other-languages" - Unlocalized languages, their names taken from that ISO-6393 library in English. I think most of those people are at least bilingual with second language from upper groups. I put it as "Rear Languages". But maybe it's better to united it with the previous group under "Other Languages" while preserving the order.
- "unknown-languages": "Unrecognized Languages" - languages, that their codes can't be detected neither with standard Internationalization API

What do you think?

zhuiks:
Ok, this is almost in place! I extracted all the steps to separate components and added Update Repository Data step (that should be first, especially when it's a new installation). Update repository step can be combined with initial internet usage warning (warning first).
Obviously, intro paragraph for each step needs to be updated. @tobias-klein  feel free to jump in. 

Please see what else I plan to finish in the first post.

tobias-klein:
Thanks for all the work, @zhuiks. I will have a detailed look during the weekend.

zhuiks:
@tobias-klein I've updated intro locale strings (for now only for English locale) to make it more clear with the new steps order. 
Please see the d711d76 commit to see what exactly was updated. I kept most of the keys the same. Should I introduce the new one instead?

I still need to fix/refactor the modules step with displaying details about the module. I'm thinking about using this emoji :information_source: instead of clicking the code link

tobias-klein:
We should try to finalize the locale updates first, so that translators have some time for their work.

tobias-klein:
I think this is not fully optimized for night mode yet?

image

tobias-klein:
@zhuiks 
How about still showing the languages of "Used and System Languages" under the other groups as well?
I was looking at Bible Languages and missed Ancient Greek. But then I noticed that this shows up under "Used Languages". I think it would have been nice to still see this under Bible Languages even though it is also in the other group.

tobias-klein:
How about coming back to a "three column layout" for the language list? I think we are wasting a bit of screen space like that and the list still gets unnecessarily long.

image

tobias-klein:
I would prefer the previously used horizontal layout for this master detail view. You hardly see any details for the selected module this way and then you have keep scrolling ... and we are not utilizing the screen space in the best way.

image

tobias-klein:
> So I guess it's ok to duplicate checkboxes. And by checking one, another one checks automatically.

Yes!

> The main reason for two column layout were few of the long language names:
> Знімок екрану з 2021-06-29 09-47-46
> But I will deal with this by other means.

OK! I thought there was a width limitation before ... which helped to have three columns.

zhuiks:
@tobias-klein would you mind checking cucumber tests on your machine.
They are passing for me, but are failing here. I can't pin-point the problem.

I think I'm done with refactoring. There are language search and few corrections left. But all code structure is there, so feel free to check it.

tobias-klein:
Thank you so much, @zhuiks! This was a lot of work and I really like the results! I will be doing some code review during the weekend. First ... here is some additional feedback regarding the UI.

1) I think we need to think about a responsive dialog size. On desktops the height should definitely be higher. At the moment I have the impression that the height of the dialog is too small. We could even think about increasing the width. At the end of the day it needs to scale from the 7" tablet screen to the full-hd desktop screen. On the desktop we may consider going almost full-window-size with the size of the dialog.
2) I would think that the display of the version and source repo may not matter all that much to the average user. It adds a lot of visual "noise" and I feel like it makes the list harder to read. We could either try to come up with a way to reduce the "noise level" or maybe make the display of that information optional / on demand? We may also consider to put source repo and version in the module info listing right on top?
image
3) The underline --- under the i button is unnecessary I think.
4) We got a new bug report #390 related to the module assistant. Can you check whether this can be resolved in the context of this PR?

tobias-klein:
> @tobias-klein would you mind checking cucumber tests on your machine.
> They are passing for me, but are failing here. I can't pin-point the problem.

The tests are also passing on my machine (both full-test and headless-test). I will have a look to see what could be wrong on GitHub actions.

tobias-klein:
> > @tobias-klein would you mind checking cucumber tests on your machine.
> > They are passing for me, but are failing here. I can't pin-point the problem.
> 
> The tests are also passing on my machine (both full-test and headless-test). I will have a look to see what could be wrong on GitHub actions.

The reason why the tests were passing locally is that the SWORD repo data was not deleted initially. So locally, repos were already existing at the point when the languages were listed. However, when you start off freshly, that's not the case. I now fixed the issue by adding a loop in `step_languages.getAvailableLanguagesFromRepos` that waits for the repos to be existing before it proceeds to populate the language list. We still need to test how this whole thing behaves if there are internet connection issues.

I also added #391, so that we can avoid this inconsistency between local and GitHub test runs in the future.

zhuiks:
> The reason why the tests were passing locally is that the SWORD repo data was not deleted initially by the Cucumber test suite. So locally, repos were already existing at the point when the languages were listed. However, when you start off freshly, that's not the case. I now fixed the issue by adding a loop in `step_languages.getAvailableLanguagesFromRepos` that waits for the repos to be existing before it proceeds to populate the language list. We still need to test how this whole thing behaves if there are internet connection issues.
> 
> I also added #391, so that we can avoid this inconsistency between local and GitHub test runs in the future.


Thanks for checking and fixing!

tobias-klein:
Thank you for the updates! The distance between the Bible modules helps to make it less "busy".

However, I feel like the vertical height of the Bible module list is still too short.

Here are the things we could do to gain some more vertical screen space:

1) Somehow compress the repo update box a bit more vertically. We could try to come up with a compressed version of the text ... or we only show a simplified box content for step 3, because there it is not so likely that you actually still need to refresh the repo data ... or we go back to how it was before (not show the repo update box on step 3).
2) Further increase the vertical dialog size.

image

app/frontend/components/module_assistant/assistant_checkbox.js
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    this._checked = this.checkbox.checked;
+    if (this._checked) {
+      this.setAttribute('checked', '');
+    } else {
+      this.removeAttribute('checked');
+    }
+
+    this.dispatchEvent(new CustomEvent("itemSelected", {
```
tobias-klein, line 180: `itemSelected` only matches the usecase when you select, but you can also deselect, so I think something more neutral would make more sense. I think I would opt for `change`, since that is also used by standard HTML elements.

app/frontend/components/module_assistant/assistant_helper.js
```diff
@@ -0,0 +1,145 @@
...
+
+module.exports.localize = function(container) {
+  container.querySelectorAll('[i18n]').forEach(element => {
+    element.innerHTML = i18n.t(element.getAttribute('i18n'));
+  });
+};
+
+module.exports.localizeContainer = function(container) {
+  var moduleTypeText = "";
+  const moduleType = assistantController.get('moduleType');
```
tobias-klein, line 135: I feel like this helper module should be purely functional with no additional dependency to the `assistantController`. I would put the `moduleType` as parameter instead and let the caller worry about getting the right value.

app/frontend/components/module_assistant/assistant_steps_add.js
```diff
@@ -0,0 +1,251 @@
...
+`;
+
+// const UPDATE_REPOSITORIES_INDEX = 0;
+const LANGUAGES_INDEX = 0;
+const REPOSITORIES_INDEX = LANGUAGES_INDEX + 1;
+const MODULES_INDEX = REPOSITORIES_INDEX + 1;
+const INSTALL_INDEX = MODULES_INDEX + 1;
+
+
+class AssistantStepsAdd extends HTMLElement {
```
tobias-klein, line 88: I would suggest to add the term "Module", so rename it to `AssistantStepsAddModule`. Let's make it crystal clear what this is all about.

app/frontend/components/module_assistant/module_assistant.js
```diff
@@ -0,0 +1,228 @@
...
+    const appContainerWidth = $(window).width() - 10;
+    var wizardWidth = null;
+
+    if (appContainerWidth < 1100) {
+      wizardWidth = appContainerWidth;
+    } else {
+      wizardWidth = 1100;
+    }
+
+    const wizardHeight = $(window).height() * 0.75;
```
tobias-klein, line 149: The ModuleAssistant used to be called "ModuleWizard". I got rid of the term "wizard" some time ago, I would suggest to also replace it here (with assistant).

app/frontend/components/module_assistant/step_install.js
```diff
@@ -0,0 +1,334 @@
...
+      if (!moduleInstalled) {
+        this.querySelector('#module-install-progress-msg').innerHTML = '';
+        await ipcNsi.installModule(moduleCode, progress => this._handleModuleInstallProgress(progress));
+      }
+
+      // FIXME
+      // Sleep a bit after installation. This is actually a hack to prevent
+      // a "white screen error" right after module installation. The exact reason
+      // for that error is unclear, but the sleeping prevents it.
+      // await sleep(100);
```
tobias-klein, line 168: If this is not needed any more we should remove it.

app/frontend/components/module_assistant/step_languages.js
```diff
@@ -0,0 +1,286 @@
...
+
+  var appSystemLanguageCodes = new Set([i18nController.getLocale(), i18nController.getSystemLocale(), ...(await getInstalledLanguages())]);
+  var bibleLanguages = new Set(['grc', 'hbo']);
+  var mostSpeakingLanguages = new Set(['en', 'zh', 'hi', 'es', 'ar', 'bn', 'fr', 'ru', 'pt', 'ur']); // source: https://en.wikipedia.org/wiki/List_of_languages_by_total_number_of_speakers
+  const historicalLanguageTypes = new Set(['ancient', 'extinct', 'historical']);
+
+  var appSystemLanguages = new Map();
+  var languages = {
+    'bible-languages': new Map(),
+    'most-speaking-languages': new Map(),
```
tobias-klein, line 212: most-spoken-languages

app/frontend/components/module_assistant/step_languages.js
```diff
@@ -0,0 +1,286 @@
...
+    'unknown-languages': new Map(),
+  };
+  
+  var allLanguages = new Map();
+  var repositories = [];
+
+  while (repositories.length < 1) {
+    repositories = await assistantController.get('allRepositories');
+    await sleep(100);
+    // FIXME: Add some sort of timeout to handle situation when repositories are never populated correctly.
```
tobias-klein, line 226: To be addressed before merge

app/frontend/components/module_assistant/step_modules.js
```diff
@@ -0,0 +1,269 @@
...
+    const loadingIndicator = moduleInfo.querySelector('loading-indicator');
+
+    moduleInfoContent.innerHTML = '';
+    loadingIndicator.show();
+
+    setTimeout(async () => {
+      const swordModuleHelper = require('../../helpers/sword_module_helper.js');
+      moduleInfoContent.innerHTML = await swordModuleHelper.getModuleInfo(moduleCode, true);
+      loadingIndicator.hide();
+    }, 200);
```
tobias-klein, line 263: We could try to reduce the timeout to 100ms.

app/frontend/components/module_assistant/step_update_repositories.js
```diff
@@ -0,0 +1,186 @@
...
+    <div id="repo-update-progress-bar" class="progress-bar">
+      <div class="progress-label" i18n="module-assistant.updating"></div>
+    </div>
+  </div>  
+  <p id="update-failed" style="display: none" i18n="module-assistant.update-repository-data-failed"></p>
+
+</section>
+`;
+
+class StepUpdateRepositories extends HTMLElement {
```
tobias-klein, line 64: This is technically not a "step" in the workflow like the others? Maybe we should remove the term "Step" from this component?

app/frontend/helpers/ui_helper.js
```diff
@@ -23,18 +23,18 @@ class UiHelper {
...
-  configureButtonStyles(context = null) {
+  configureButtonStyles(context=null) {
     if (context == null) {
       context = document;
-    } else {
-      var context = document.querySelector(context);
-    }
+    } else if (typeof context === 'string') {
+      context = document.querySelector(context);
+    }
```
tobias-klein, line 32: Now the `else` case is not accounted for anymore ... we could consider throwing an exception in the else case.

css/main.css
```diff
@@ -45,14 +56,15 @@ body {
...
   overflow: hidden;
   touch-action: none;
+  background-color: var(--background-color);
 }
 
-body.darkmode--activated {
+/* body.darkmode--activated {
   background-color: #1e1e1e !important;
-}
+} */
```
tobias-klein, line 64: Let's remove it if this is not needed anymore.

features/001_module_installation.feature
```diff
@@ -26,8 +26,9 @@ Feature: Module installation
   Scenario: ASV installation
     Given I open the module installation assistant
     And I choose to add translations
-    And I select the CrossWire repository
+    And I check repository data update date
```
tobias-klein, line 29: "And the repository update date matches today"

tobias-klein (commented):
Once more ... good job! I like the way you "sliced the elephant"! Just some minor feedback!

locales/en/translation.json
```diff
@@ -75,7 +76,13 @@
...
     "unlock-key": "Unlock key",
     "enter-unlock-key": "{{moduleId}} is license protected. Please enter the unlock key!",
-    "unlock-failed": "Unlock failed! Please enter the correct unlock key."
+    "unlock-failed": "Unlock failed! Please enter the correct unlock key.",
+    "app-system-languages": "Used and System Languages",
+    "bible-languages": "Bible Languages",
+    "most-speaking-languages": "Most Spoken Languages",
+    "historical-languages": "Historical Languages",
+    "other-languages": "Other Languages",
+    "search-result": "Matches Found"
```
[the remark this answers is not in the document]
zhuiks: Question: for locale strings that changed but have the same key (`module-assistant.pick-languages-from-repos`, `module-assistant.repo-selection-info-text`, etc.) should I just put new English text for all locales or is it better to use new keys instead?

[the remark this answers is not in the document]
tobias-klein: I would suggest to copy & paste the new English text.

zhuiks:
> However, I feel like the vertical height of the Bible module list is still too short.

How about this. There needs to be some style adjustments still. But the idea is there.

tobias-klein:
I like the way how you moved most of the text about the repos to the first dialog page. Now we only need to squeeze the remaining box size a bit more, I think it could be shrink almost 50% still.

Regarding the popup info inside the existing scroll box:
1) I personally do not like the nested scrollable box. If we do it like this I'd rather just scroll all the way and not have an additional scrollable box inside the overall scrollable box.
2) I don't like that the info box sometimes appears above the info button and sometimes below it. I think I'd much rather have it consistently show up in the same place.

zhuiks:
1. Update box. I removed "Update takes some time..." text over progress bar after you press update. I think It's worth to move it into the "steps" area for the first two steps, and don't show it for the last two steps:
Знімок екрану з 2021-07-06 22-02-58

2. Module info. Playing with it, I'm half-half about returning to the initial layout with dedicated info area on the right. It just should be hidden initially and revealed on info button click. If we stick with the current "updated" layout, the info button needs to be adjusted to have active state and allow for multiple info boxes to be open. Currently, opening one will close another one, which gives scrolling out of view side effects.
Знімок екрану з 2021-07-06 22-02-16

tobias-klein:
> 2\. Playing with it, I'm half-half about returning to the initial layout with dedicated info area on the right. It just should be hidden initially and revealed on info button click.

I would opt for that (reveal initially and show on info button click).

app/frontend/components/generic/fuzzy_search.js
```diff
@@ -0,0 +1,173 @@
...
+
+    if (!this._fuse) {
+      return;
+    }
+
+    const value = event.target.value.trim();
+    this._searchTimeout = setTimeout(() => {
+      const searchResult = value.length < this.minLength ? [] : this._fuse.search(value);
+
+      this.dispatchEvent(new CustomEvent("searchResultReady", {
```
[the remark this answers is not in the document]
zhuiks: done

app/frontend/components/generic/fuzzy_search.js
```diff
@@ -0,0 +1,169 @@
...
+</form>
+`;
+
+/**
+ * @module FuzzySearch
+ * component displays search input field and triggers 'searchResultReady'
+ * @example
+ * <fuzzy-search min-length="2"></fuzzy-search>
+ * @category Component
+ */
```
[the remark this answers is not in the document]
zhuiks: done

package.json
```diff
@@ -71,6 +71,7 @@
     "electron-window-state": "^5.0.3",
     "emoji-button": "github:zhuiks/emoji-button",
     "fs-extra": "^9.0.1",
+    "fuse.js": "^6.4.6",
```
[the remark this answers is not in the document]
zhuiks: done

app/frontend/components/module_assistant/assistant_checkbox.js
zhuiks: The catching of this event is done by step web components (it bubules up). So I will opt for unique name so random input or checkboxes would not trigger the handler. `itemChanged`

locales/en/translation.json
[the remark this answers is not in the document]
zhuiks: I went through all the texts for module installation with my father in law. He is semi-technical. I got more updates than I initially expected. But in the end it turned out to be more clear for the common user. I realized that it's important to avoid term "module" completely. I added quite a few new keys to reflect the updated text.

zhuiks:
@tobias-klein I went a bit crazy with i18n strings. There was a lot of splitting, renaming and reorganizing in module-assistant section. I heavily consulted with my wife and father-in-law :) I think it would be much more clear for the common user this way. Although maybe I opened a can of worms with translator's opinions. I only added new strings to German locale. If you approve this changes, I'll add new strings to all other locales.

Regarding #390. If there is no internet connection, the error message is correctly shown. But if connection disappears while updating - no error is thrown on `nsi` level. 

I was thinking to preload languages if repositories are available. But it's important not to go online on the first module-assistant view (it would neglect the internet usage warning). But step_languages component is not constructed until the 'add-button' is clicked because of jQuery Steps plugin needs to be loaded first and do its tricks with DOM. When we upgrade to something else it would be possible.

tobias-klein:
> Regarding #390. If there is no internet connection, the error message is correctly shown. But if connection disappears while updating - no error is thrown on `nsi` level

It seems like the return code of updateRepositoryConfig is not correctly evaluated in assistant_controller. There won't be an exception. It's based on the return code.

This is how it is done on master at the moment:
https://github.com/ezra-bible-app/ezra-bible-app/blob/26e3d4485287410820af606cca89ced714c3ce77/app/frontend/components/module_assistant/install_module_assistant.js#L162

I'll do some more testing and also get back regarding the locale changes.

tobias-klein:
> It seems like the return code of updateRepositoryConfig is not correctly evaluated in assistant_controller. There won't be an exception. It's based on the return code.

This turned out as a bug in node-sword-interface. The return codes from SWORD were not correctly "passed down" to the caller. This is fixed now.

tobias-klein:
I updated the following locale texts:
- Bible translations are ~~published~~ by the CrossWire Bible Society => Bible translations are **distributed** by the CrossWire Bible Society. (CrossWire is *not* a Publisher like Lockman Foundation, it just distributes the modules)
- Used and System Languages => **Existing Languages** and System Languages (Easier to understand that way)

Regarding the following sentence on the initial dialog:
"You may select languages and particular repositories from which you will install Bible translations."

After thinking about it for some time, I would actually suggest to remove the entire sentence. The assistant with its steps is self-explanatory I think and with this additional text the dialog gets quite text-heavy.

tobias-klein:
I found a bug:
When you pick a language where the corresponding module(s) are only available from one particular repository, then the repository should be auto-selected based on the language choice. That is not the case at this point.

tobias-klein:
I found another bug:
When you install dictionaries the texts in the first dialog are still talking about "Bible translations".

image

zhuiks:
> The remaining feedback I have is ... can we show the error message a little while longer? My impression is that it is shown too shortly.

How about showing extended error in the Language list or Repository list section (in place of "Loading languages..."/"Loading repositories..." phrase. It would be quite easy to implement as the status/result from [`updateRepositoryConfig`](https://github.com/ezra-project/ezra-project/blob/fb4460cd62e113f2eb99494f1c9b4bcc05206dae/app/frontend/components/module_assistant/assistant_controller.js#L142) is just passed down to subscribers.

Thanks for checking/updating the text locale.
> Regarding the following sentence on the initial dialog:
"You may select languages and particular repositories from which you will install Bible translations."

> After thinking about it for some time, I would actually suggest to remove the entire sentence. The assistant with its steps is self-explanatory I think and with this additional text the dialog gets quite text-heavy. If we keep it I would suggest to change it, because it is actually not "may" but rather "must" and the purpose of the sentence is a bit vague the way it is written. I would say something like "To find the right Bible translations, you first need to select preferred languages and after that repositories from which you can install Bible translations."

My initial thought was to give the right expectations to a user. Seems like the concept of selecting repositories is a bit confusing from my conversation with a few test users. But I would agree: let's remove this sentence.

> I found a bug:
When you pick a language where the corresponding module(s) are only available from one particular repository, then the repository should be auto-selected based on the language choice. That is not the case at this point.

I want to make sure I understand this correctly. If I select "Church Slavic" language and it's only available from CrossWire - CrossWire should be auto selected. Or the same with "Adyghe" language, for instance, - IBT should be automaticly selected as it's the only repository where "Adyghe" is available. 
Is this  correct?

tobias-klein:
> How about showing extended error in the Language list or Repository list section (in place of "Loading languages..."/"Loading repositories..." phrase. It would be quite easy to implement as the status/result from [`updateRepositoryConfig`](https://github.com/ezra-project/ezra-project/blob/fb4460cd62e113f2eb99494f1c9b4bcc05206dae/app/frontend/components/module_assistant/assistant_controller.js#L142) is just passed down to subscribers.

Sure!

> I want to make sure I understand this correctly. If I select "Church Slavic" language and it's only available from CrossWire - CrossWire should be auto selected. Or the same with "Adyghe" language, for instance, - IBT should be automaticly selected as it's the only repository where "Adyghe" is available.
> Is this correct?

Yes!

zhuiks:
Another thought about locale texts.
It seems that there is a struggle to translate "Bible translation(s)" and "Dictionary(ies)" in the proper case. English grammar doesn't have any cases for nouns but, for instance, Slavic languages have 6-7 cases. I was looking how it's done in [fluent](https://projectfluent.org/fluent/guide/terms.html), and it seems like we can do it with i18next with [context](https://www.i18next.com/translation-function/context) and [nesting](https://www.i18next.com/translation-function/nesting#passing-options-to-nestings). Although it would bring extra complexity to certain locale strings, we can remove duplication of strings where the difference is just "translation" or "dictionary" and give more flexibility to translators to use different cases.

Probably this can be a separate issue/PR

tobias-klein:
> Probably this can be a separate issue/PR. But it needs to be done before translators get to the new locale strings

I leave it to you whether you would like to still handle that change in this PR or in the context of another new issue. I agree that it makes sense.

tobias-klein:
Just some feedback regarding the performance of the list language page. Note that the measurements were done on a Core i7, so you can expect much longer execution times on a slow device.

This is before (master):
image

And after (this branch):
image

If you look into the details you find this:
image

The two functions that are taking a big toll are:
* `sortByText`
* `getLanguageDetails`

Note that I already optimized `findLanguage` which gets called by `getLanguageDetails`.
Overall this is factor three slower than before the refactoring ... I wonder whether additional optimizations are still possible.

zhuiks:
> > Probably this can be a separate issue/PR. But it needs to be done before translators get to the new locale strings
> 
> I leave it to you whether you would like to still handle that change in this PR or in the context of another new issue. I agree that it makes sense.

I decided to implement it here to copy new locale strings only once.

zhuiks:
> The two functions that are taking a big toll are:
> * `sortByText`
> * `getLanguageDetails`
> 
> Note that I already optimized `findLanguage` which gets called by `getLanguageDetails`.
> Overall this is factor three slower than before the refactoring ... I wonder whether additional optimizations are still possible.



`getLanguageDetails` can be farther optimized with #362. We just need to build a proper language code map with ISO639 and SWORD data in advance. Similar to what you did with `findLanguage` but for all language codes.

I optimized and simplified sorting. Although I still need to adjust other step components with the new sorting approach.

Seems like currently the bottle neck is with layout/reflow. I'll see if I can do anything with it.

Another idea is to move all calculation logic into controller and make all step components as simple UI wrappers. That way I think it would be possible to trigger language compilation on opening the dialog if repository data is already available

tobias-klein:
I just tested the language listing performance on my Surface Go tablet. Before it was six seconds, now it is 18 seconds. That confirms factor three mentioned above. I have not tested this on Android yet, but I suppose it is even slower. I think we need to check how to further improve the performance ... I fear this is too slow.

I think generally speaking the performance issue is that individual web components are also appended to the DOM individually. When you do a batch action (first consolidating the html and then one innerHTML assignment) it is usually much faster. But I guess this technique cannot be used with the web components approach that we have here.

zhuiks:
> I just tested the language listing performance on my Surface Go tablet. Before it was six seconds, now it is 18 seconds. That confirms factor three mentioned above. I have not tested this on Android yet, but I suppose it is even slower. I think we need to check how to further improve the performance ... I fear this is too slow.

Have you tried with the latest commits?

I also notice that the first opening of module assistant seems to be much slower than all the next openings. `getRepoLanguages` seems to be the reason for longer wait. 

That's what I'm getting with CPU throttling 6x slowdown on i7 and my take on it:
Знімок екрану з 2021-07-16 08-24-33

The first fade in animation section on the screenshot is actually waiting for `getAllLanguageModuleCount()`. Animation itself is a very small "bump" with browser paint inside it.

Recalculate Styles and Layout caused by >1000 AssistantCheckbox component insertion seems to be the bottleneck.

> I think generally speaking the performance issue is that individual web components are also appended to the DOM individually. When you do a batch action (first consolidating the html and then one innerHTML assignment) it is usually much faster. But I guess this technique cannot be used with the web components approach that we have here.

I tried to generate one big HTML string with all language sections together and insert it with `innerHTML`. It seems to be similar or longer in performance, with the "layout" task being the longest:
Знімок екрану з 2021-07-16 10-40-43

zhuiks:
I got the bad one with 6x slowdown (I think the image before was captured without slowdown):
Знімок екрану з 2021-07-16 12-08-43

And this what I get currently in master (with 6x slowdown):
Знімок екрану з 2021-07-16 13-05-39

zhuiks:
@tobias-klein I think I boiled it down to the one problematic [line step_languages#216](https://github.com/ezra-project/ezra-project/blob/57baf8a9f315cee038e6a60663b9c2be6efdc3b7/app/frontend/components/module_assistant/step_languages.js#L216):
```
    this._allLanguages.appendChild(containerLongList);
``` 
You can comment it and see how everything gets quite fast. I think I'll try to use [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) to append elements dynamically

zhuiks:
I moved the long tasks to `requestIdleCallback`. Intersection Observer API recommended being used with something fast.

Seems like timing to the user interaction is comparable to what is in the master now.

Let me know if there is anything else to be fixed, otherwise it can be merged!

.eslintrc.json
```diff
@@ -0,0 +1,46 @@
...
+        "ipcDb": "readonly",
+        "ipcGeneral": "readonly",
+        "app_controller": "readonly",
+        "tags_controller": "readonly",
+        "theme_controller": "readonly",
+        "uiHelper": "readonly",
+        "i18n": "readonly",
+        "platformHelper": "readonly"
+    },
+    "extends": "eslint:recommended",
```
zhuiks, line 27: @tobias-klein 
Please, note that this PR will introduce eslint rules a bit extended rules from #320 
Rules are based on ESLint recommendations

tobias-klein:
Thank you for the optimizations! The (perceived) performance is much better now! :)
The only change that I would still suggest for the language listing is to show the **loading indicator** until the very end (only hide it when all languages have been fully rendered).

.eslintrc.json
tobias-klein: Ok, cool :)

locales/de/translation.json
```diff
@@ -32,50 +32,80 @@
...
+      "module-display-preferences": "Filter preferences:",
+      "module-with": "{{module_type, capitalize}} with {{feature_type, instrumental}}",
+      "module-already-installed": "This {{module_type, singular}} is already installed",
+      "show-module-info":"Show info",
+      "show-detailed-module-info": "Press on the information icon for more details about the {{module_type, singular}}."
+    },
+    "step-install": {
+      "installing-modules": "Installing selected {{module_type, accusative}}",
+      "installing": "Installiere",
+      "it-takes-time-to-install": "Note that it takes some time for each {{module_type, singular-genitive}} to be downloaded and then installed.",
```
tobias-klein, line 93: In my mind, this is accusative ... (the module type is the object in this sentence).

tobias-klein:
I am planning to update the German locale texts on this branch before we merge ... this will be a good test whether the new locale structure works well for translators. I'll do that this week.

tobias-klein:
> Thank you for the optimizations! The (perceived) performance is much better now! :)
> The only change that I would still suggest for the language listing is to show the **loading indicator** until the very end (only hide it when all languages have been fully rendered), because currently it creates the impression that it is done with loading even though that is not the case.

Thanks for the updates! This is nice now :)

tobias-klein:
I think it would be visually more pleasing if the info button would be vertically aligned (in a column). But I guess to make that happen we would need a table? Or is it possible with pure CSS?

image

We can also keep it like it is and turn this one into a new issue?

tobias-klein:
I have another performance feedback. When you click the fuzzy search button in the Bible translation language list, it takes quite a bit until the search field shows and you kind of notice a "lagging effect". Then I did a performance measurement and it looks like this (on my Core i7):
image

It seems like significant reflow happens when the search field is shown. Maybe we can use `visibility: hidden` instead of `display: none` (I didn't actually check the code for this), to avoid the reflow?

zhuiks:
I've fixed info button aligning.

Regarding search and language list. For now, I just turned off search box animations. The bottleneck is the reflow/layout with large amount of nodes. I think it's similar to the BibleBrowser's long chapters delays. I'll keep researching and I hope to find a solution to both.

zhuiks:
@tobias-klein lately I've been noticing that sometimes updating repositories will give an error. The problem will disappear over the time. I think it might be an issue connected with one of the repositories being offline, as it was discussed in sword-developers mailing list recently.

tobias-klein:
> @tobias-klein lately I've been noticing that sometimes updating repositories will give an error. The problem will disappear over the time. I think it might be an issue connected with one of the repositories being offline, as it was discussed in sword-developers mailing list recently.

Oh, ok. Interesting. I wonder whether there's anything we can do about it. At the moment the respective node-sword-interface function fails if any of the individual repo updates failed. An alternative would be to fail with different types of exceptions or error Codes depending on what happens.

zhuiks:
> Oh, ok. Interesting. I wonder whether there's anything we can do about it. At the moment the respective node-sword-interface function fails if any of the individual repo updates failed. An alternative would be to fail with different types of exceptions or return codes depending on what happens:
> 
>     * success
> 
>     * some repos failed
> 
>     * all repos failed

Yes, if you can implement an extra status for "some repos failed" that would be helpful. In this case, we still can work with some repositories and Cucumber test won't fail.

tobias-klein:
I just tested on Android. Everything works fine ... except I discovered a new bug, but that is already in 1.1 (see #394). We can solve that separately.

@zhuiks 
Can you check the formatting of the checkbox labels? On Android the text needs to be moved to the top a little (with vertical margin?). I've been having the same issue with checkboxes in other places (tag assignment menu, tag selection menu) - so maybe you find some existing CSS solution.

image

locales/de/translation.json
tobias-klein: @zhuiks Any feedback on this one?

tobias-klein:
> Yes, if you can implement an extra status for "some repos failed" for node-sword-interface that would be helpful. In this case, we still can work with some repositories and Cucumber test won't fail.
> Maybe node-sword-interface can pass on extra object with failed repositories.

I created this issue now: #395 
We can solve it independent of this PR.

tobias-klein:
`Click on the information icon to see more details about the {{module_type, singular}}.`
This does not work well in German (and maybe even other more complex languages like French or Greek), because the _demonstrative pronoun_ (this) in front of the `module_type` depends on the gender of the `module_type`.

* the Bible translation: die Bibelübersetzung
* the dictionary: das Wörterbuch

Any suggestions on how to solve that?

zhuiks:
> `Click on the information icon to see more details about the {{module_type, singular}}.`
> This does not work well in German 

> Any suggestions on how to solve that?
> I guess one way would be to move the pronoun into the translation of module_type.

Yes, I suggest that you move the pronoun to the `module_type` translation. I updated `CONTRIBUTION.md` to explain more about that and put this case as an example. Word forms can be individual for each language. Feel free to rephrase what I wrote there. Maybe in future we can replace examples with Slovakian instead of Russian. I'm guessing the grammar of the Slovakian is quite similar to other Slavic languages. But it uses Latin characters, so it would be more universal as an example.

zhuiks:
I'll fix Android checkbox issue before merge.

tobias-klein:
I found another bug.

When you run the install module assistant the second time, it is possible to proceed from step 3 (module selection) to step 4 (installation) without actually selecting any module.

It looks like this then:
image

tobias-klein:
I created #396. Otherwise, we can merge once the above mentioned bug is fixed.

zhuiks:
> I found another bug.
> 
> When you run the install module assistant the second time, it is possible to proceed from step 3 (module selection) to step 4 (installation) without actually selecting any module. This should not be possible (It is successfully prevented when running the module assistant the first time).

That should be fixed now. It was the tricky issue with resetting jqury.steps data

tobias-klein (approved):

tobias-klein:
This has been a lot of work @zhuiks! Thank you so much :) Finally merging now!

PR #364, opened 2021-06-19
Reorder options: Move general options to the left side and layout and text options to the right side of options menu

tobias-klein:
@zhuiks
@augustin-colesnic 
I would prefer this order of the optoin groups in the options menu.
Note that there were also some changes within the options groups compared to 1.0.0.
Please have a look and give feedback.

Please approve via the GitHub review functionality if you are OK with this change.

**NEW** (1.1.0)

image

**OLD** (1.0.0)

image

### DIFF
--- a/html/display_options_menu.html
+++ b/html/display_options_menu.html
@@ -13,11 +13,19 @@
 </div>
 
 <div class="options-group">
-  <div class="options-header" i18n="bible-browser.display"></div>
+  <div class="options-header" i18n="bible-browser.bible-browser-layout"></div>
 
   <config-option id="showToolBarOption" settingsKey="showToolBar" label="bible-browser.show-tool-bar" checkedByDefault="true"></config-option>
   <config-option id="showDictionaryOption" settingsKey="showStrongs" label="bible-browser.show-dictionary" checkedByDefault="false"></config-option>
-  <config-option id="openVerseListsInNewTabOption" settingsKey="openVerseListsInNewTab" label="bible-browser.open-verse-lists-in-new-tab" checkedByDefault="false"></config-option>
+  <config-option id="useTagsColumnOption" settingsKey="useTagsColumn" label="bible-browser.use-tags-column" checkedByDefault="false"></config-option>
+  <config-option id="showBookChapterNavigationOption" settingsKey="showBookChapterNavigation" label="bible-browser.show-book-chapter-navigation" checkedByDefault="true"></config-option>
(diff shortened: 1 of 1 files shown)


zhuiks:
@tobias-klein can we put "Night mode" as a first one in **Display** section. I think that's the standard option that users would look for

html/display_options_menu.html
```diff
@@ -26,11 +34,14 @@
     <option value="open-chapters-all-books" label="bible-browser.open-chapters-all-books"></option>
   </select>
 
+  <config-option id="openVerseListsInNewTabOption" settingsKey="openVerseListsInNewTab" label="bible-browser.open-verse-lists-in-new-tab" checkedByDefault="false"></config-option>
+  <config-option id="adjustTagsNotesTextSizeOption" settingsKey="adjustTagsNotesTagSize" label="bible-browser.adjust-font-size-tags-notes" checkedByDefault="false"></config-option>
+  <config-option id="fixNotesHeightOption" settingsKey="fixNotesHeight" label="bible-browser.limit-notes-display-size" checkedByDefault="false"></config-option>
   <config-option id="useNightModeOption" settingsKey="useNightMode" label="bible-browser.use-night-mode" checkedByDefault="false" autoLoad="false"></config-option>
   <config-option id="keepScreenAwakeOption" settingsKey="keepScreenAwake" label="bible-browser.keep-screen-on" checkedByDefault="false"></config-option>
```
zhuiks, line 41: How about moving these two option as first ones in **Display** section (insert after line 28). These are standard options and users would look for them.

zhuiks (changes requested):
@tobias-klein This is good rearrangement! If we will move this to a right one column sidebar as discussed in https://github.com/ezra-bible-app/ezra-bible-app/discussions/299#discussioncomment-777004 it would be nice to keep the same new order.

html/display_options_menu.html
tobias-klein: Yes, makes sense. I just pushed that change.

tobias-klein:
@zhuiks I implemented the change you requested. Please approve the PR.

augustin-colesnic:
Hi @tobias-klein , 

I fails to see night mode on the settings.

Screenshot 2021-06-21 at 11 17 11
Screenshot 2021-06-21 at 11 25 08

And @zhuiks  I found uncomfortable the timeout that makes the settings panel disappear after I check an option...  Old behaviour was better.

zhuiks (approved):
Thanks!

tobias-klein:
@tobias-klein On macOS the night mode is controlled via system settings only. When you change it in the system settings the app will respond to that change right away.

PR #302, opened 2021-05-21
Update app layout based on CSS grid

zhuiks:
This solves #77
The idea discussed in #299

### Summary of changes:
* refactor main layout inside appContainer with CSS grid
* refactor layout inside verseListTab with CSS grid
* responsive tagList and dictionary sections inside sidePanel implemented in pure CSS (without a need to adjust sizes with javascript)
* on small screen sizes (< 1200px) sidePanel is hedden and dictionary section shows underneath BibleBrowser

PR #374, opened 2021-06-22
Android toolbar layout fix

### LINKED ISSUE #337: Android: Showing toolbar results in invalid / buggy layout
This was observed on Android. When switching on the tool bar, I got this as a result ... and this feels like a regression (I don't think this is a bug in any release).

Screenshot_2021-06-06-11-50-05-253~2

zhuiks:
fixes #337

### DIFF
--- a/css/toolbox.css
+++ b/css/toolbox.css
@@ -48,7 +48,7 @@
   height: calc(100% - 7px);
 }
 
-.with-dictionary > #tags-content {
+:not("body.Android") .with-dictionary > #tags-content {
   height: 50%;
 }
 


tobias-klein:
Mmmm... I tested this just now and I couldn't reproduce the bug anymore. However, after switching back to master I also couldn't reproduce the bug anymore. The thing is that technically the dictionary can't even be turned on with Android (the option is hidden).

tobias-klein:
I actually reproduced this again (it happens once you have opened the drop down tag list at least once) and the fix helps (to some extent). What's missing is that the vertical height of the tag list is "endless" instead of adjusted to the window height. I think this is handled in UiHelper.resizeAppContainer. I may have another look later this week and see whether I can fix this. 

image

zhuiks:
> I actually reproduced this again (it happens once you have opened the drop down tag list at least once) and the fix helps (to some extent). What's missing is that the vertical height of the tag list is "endless" instead of adjusted to the window height. I think this is handled in UiHelper.resizeAppContainer. I may have another look later this week and see whether I can fix this.

Can we put this on hold for now. I really want to finish #302 which refactor the whole layout without representing new features. My goal to adjust resizing in pure CSS without utilizing JavaScript (`UiHelper.resizeAppContainer`)

tobias-klein:
> Can we put this on hold for now. I really want to finish #302 which refactor the whole layout without representing new features. My goal to adjust resizing in pure CSS without utilizing JavaScript (`UiHelper.resizeAppContainer`)

Sure. Then let's prioritize #302 for 1.2.

tobias-klein:
@zhuiks Feel free to close this if you think everything is addressed in #302.

zhuiks:
Everything addressed in #302

PR #387, opened 2021-06-29
Emoji browserify and dynamic locale optimization

### LINKED ISSUE #369: Optimize emoji annotations and browserify
Emoji annotations can be `required` dynamically, but that should be done in a separate file which should be excluded from browserify build
See https://github.com/ezra-bible-app/ezra-bible-app/pull/226#issuecomment-830739907

zhuiks:
This closes #369 

External emoji library and locale files for it (emoji annotations) are ignored for browserify build (we don't need it in Cordova)

So it should be thoroughly tested on Android


app/frontend/helpers/emoji_helper.js
```diff
@@ -0,0 +1,43 @@
...
+ * It should not be included in browserify build as it doesn't support dynamic requires
+ * and there is no need for emoji picker under Android or other mobile platform
+ * @module emojiHelper
+ * @category Utility
+ */
+
+const LIB_PATH = 'emoji-button/dist/';
+
+module.exports.getEmojiButtonLib = function() {
+  const { EmojiButton } = require(LIB_PATH + '/index.cjs');
```
tobias-klein, line 31: I don't think Browserify works with variables in require statements. We had issues with that before. What Browserify does is essentially pre-load the content of the require at bundling time and essentially put it in a large string inside the JS bundle.

tobias-klein (changes requested):

app/frontend/helpers/emoji_helper.js
zhuiks: The whole purpose of this PR is to move all Browserify incompatible or unneeded in Cordova code to one file and ignore it in Browserify build (see updates to `packaje.json`) 
Sorry, if I didn't communicate it clearly initially.

tobias-klein: You're right!

tobias-klein:
I tested:
1) Startup / No JavaScript issues
2) Switching the UI language / No JavaScript issues
3) Another startup with changed language / No JavaScript issues

PR #397, opened 2021-07-27
Install assistant info button

### LINKED ISSUE #396: Install assistant: Align module info button horizontally in column
I think it would be visually more pleasing if the info button would be vertically aligned (in a column). But I guess to make that happen we would need a table? Or is it possible with pure CSS?
> 
> image

zhuiks:
Fixes #396


tobias-klein (approved):

tobias-klein:
Thanks!

Issue #389, opened 2021-06-30
Android: implement dictionary word lookup for touch screens


zhuiks:
Long touch (hold) on the word seems logical addition to the `Shift`+ `mouse over` for dictionary article lookup

tobias-klein:
Yes, nice idea. Let's give this a try - I am curious about the results :)

tobias-klein:
With that we could then also finally enable the dictionary box on Android!

PR #398, opened 2021-07-31
Feature strong long press

zhuiks:
This solves #389 

It should be merged after #302


tobias-klein:
This is fun! Thank you so much for this @zhuiks! Excellent work.

I reduced the long-press delay to 500ms - that is actually sufficient when using it on a tablet!

tobias-klein (approved):

PR #401, opened 2021-08-01
Disable create/rename tag button if tag title exists or is empty

tobias-klein:
This introduces basic validation of the tag title when creating new tags and when renaming existing tags.

* The create/update tag button is disabled when the tag is empty or if it already exists.
* The button gets enabled again once it is not empty and does not exist yet.

### DIFF
--- a/app/frontend/controllers/tags_controller.js
+++ b/app/frontend/controllers/tags_controller.js
@@ -106,19 +106,52 @@ class TagsController {
     new_standard_tag_dlg_options.buttons[i18n.t("general.cancel")] = function() {
       $(this).dialog("close");
     };
-    new_standard_tag_dlg_options.buttons[i18n.t("tags.create-tag")] = function() {
-      tags_controller.saveNewTag(this, "standard");
+    new_standard_tag_dlg_options.buttons[i18n.t("tags.create-tag")] = {
+      id: 'create-tag-button',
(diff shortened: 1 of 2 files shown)


app/frontend/controllers/tags_controller.js
```diff
@@ -106,21 +106,54 @@ class TagsController {
...
         tags_controller.saveNewTag(event, "standard");
       }
+    }).on("keyup", async (event) => {
+      var tag_title = $('#new-standard-tag-title-input').val();
+      await this.updateButtonStateBasedOnTagTitleValidation(tag_title, 'create-tag-button');
     });
   }
 
+  async updateButtonStateBasedOnTagTitleValidation(tagTitle, buttonId) {
+    var tag_existing = await this.tagExists(tagTitle);
```
zhuiks, line 139: Would `tagExists` be fast enough for each `keyup` event?

zhuiks (commented):
Everything looks fine.
I have one question and one comment

app/frontend/controllers/tags_controller.js
tobias-klein: On the desktop - yes!
On Android - I still need to test.

tobias-klein: Sufficiently fast on Android as well - just tested.

PR #403, opened 2021-08-02
Fix commit info

### LINKED ISSUE #388: Build process assumes Git clone, doesn't work from archive
Something in the last release cycle regressed and assumes all sources are Git clones.

There is a special Git attribute you can use to stuff git commit hash information automatically in `git archive` generated tarballs, but whatever is calling Git is trying to do so directly:

```
npm ERR! code ENOENT
npm ERR! syscall spawn git
npm ERR! path git
npm ERR! errno -2
npm ERR! enoent spawn git ENOENT
npm ERR! enoent This is related to npm not being able to find a file.
npm ERR! enoent
```

Interestingly it doesn't fail if it can't find the relevant Git repository, it just fails if `git` is not available. A assume there is a fallback somewhere for handling archive sources vs. clones, but the logic is wrong and shouldn't assume the availability of Git tooling if it isn't actually required to build.

See also #23 which is coming along but not quite there yet.

tobias-klein:
This should fix #388.


alerque (commented):
Looking at the code I *think* this makes sense, but a couple caveats...

* It's not quite clear to me how one would inject commit info into a build system that wasn't a clone. Won't this result in builds of stable tags result in a detailed string if built from a Git clone and an empty string if built from an archive? It seems like that is undesirable and either tagged releases should *always* avoid the extra info or the extra info should be stuffed in the archive using Git meta properties.
* It's just a bit tricky to test this situation. I'll try to run some tests on archives built from this PR. I'll mark the review status based on those results when I get anything.

tobias-klein:
> It's not quite clear to me how one would inject commit info into a build system that wasn't a clone. Won't this result in builds of stable tags result in a detailed string if built from a Git clone and an empty string if built from an archive? It seems like that is undesirable and either tagged releases should _always_ avoid the extra info or the extra info should be stuffed in the archive using Git meta properties.

Yes, if you build from an archive you would not be able to use the `git-to-json` package. Technically - if you are somehow generating that information in a different way / from a different source than the archive itself, then you could just put that file (`app/commit_info.js`) into the folder structure before generating the package.

tobias-klein:
> It's just a bit tricky to test this situation. I'll try to run some tests on archives built from this PR. I'll mark the review status based on those results when I get anything.

@alerque 
Any feedback?

tobias-klein:
I am merging this for now. If it still causes issues, please open a new issue @alerque.

Issue #362, opened 2021-06-17
Use SWORD locale data as a fallback for Javascript i18n API


zhuiks:
As a fallback when Javascript i18n API not available or language code is not ISO-639.1 or ISO-639.2 code, it would be nice to use data from SWORD `locales.d/locales.conf` for language list in module assistant.

zhuiks:
I think SWORD locale config should be compiled as a separate npm task to JSON to be used inside `localeMapper.getLanguageDetails` similar to `'iso-639-3'` library

tobias-klein:
I guess the question is whether the benefit of this change justifies the effort ... :)

zhuiks:
I think it's a good way to test out the setup for pulling data from SWORD to JSON which we can utilize for #304

PR #406, opened 2021-08-07
Pre parse SWORD languages

zhuiks:
This is for #362.
Instead of using `iso-639-3` package during runtime, it pre parses it based on [SWORD locales.conf](https://crosswire.org/svn/sword/trunk/locales.d/locales.conf).

That will improve language look up performance.


zhuiks:
Regarding performance.
This is with the current commits in the PR on i7 with 6x slowdown:
Знімок екрану з 2021-08-06 18-58-23

However, if we use the same code but for all `iso-639-3` "indexed" languages, performance drops:
Знімок екрану з 2021-08-06 19-05-11


Although, layout is still the bottleneck.

scripts/sword_locales_to_json.js
```diff
@@ -0,0 +1,208 @@
+const fetchline = require('nodefetchline');
+const fs = require('fs');
+const iso6393 = require('iso-639-3');
+const i18nController = require('../app/frontend/controllers/i18n_controller.js');
+
+
+const SWORD_LANGUAGES_URL = "https://crosswire.org/svn/sword/trunk/locales.d/locales.conf";
```
tobias-klein, line 7: Can we instead use the files that are part of node-sword-interface?
Always fetching the latest files from CrossWire leads to un-deterministic behavior, because it is not using a fixed version of SWORD.

zhuiks: > Can we instead use the files that are part of node-sword-interface?

Yes, that makes sense.

> Also I am wondering whether we should move this script to node-sword-interface in that case.
We would then simply generate an updated json file each time when there are changes in a new node-sword-interface version.

It's a nice idea. I was wondering myself if something like that should be a part of the backend. But it's nice to have a separate `json` file with all language details. 
But if we move it to `node-sword-interface` how would you suggest pulling a list of all available languages from `i18n_controller`?

Also, please note that the script is benefitting from node 14 that implements the "new" Intl API

zhuiks:
There is an edge case, when a user of Ezra creates a new SWORD module with the language that is not in the list. I think his/her expectation would be to find that language in the list. But new language details would come only after regenerating languages.json with next Ezra update. Although he/she will see at least the new language code in the Installation assistant language list.

As for now, I think, this behavior would be okay with current user database. And a little gain on performance + simpler code has a higher priority.

tobias-klein:
Thank you for this!

I would suggest to:
* Move the new scripts underneith `build_scripts`.
* Turn the action `convert-sword-locales` into a `postinstall` action in `package.json`.

tobias-klein:
@zhuiks 
So ... which part of this pull request is still valid now that the script is moved to node-sword-interface?

zhuiks:
> @zhuiks
> So ... which part of this pull request is still valid now that the script is moved to node-sword-interface?

As it was mentioned in the comment in https://github.com/ezra-bible-app/node-sword-interface/blob/9e5d76628eb11bb047871d3f8c81ee96b104b1f9/scripts/sword_locales_to_json.mjs#L32 because this script is taking available locale list from this Ezra Bible App repository, it would be not a good practice to make the loop-back dependency from Node Sword Interface. So maybe this script should be back in this repository.

I'm putting all connected PR in draft until we will come to conclusion on this here

tobias-klein:
> > @zhuiks
> > So ... which part of this pull request is still valid now that the script is moved to node-sword-interface?
> 
> As it was mentioned in the comment in https://github.com/ezra-bible-app/node-sword-interface/blob/9e5d76628eb11bb047871d3f8c81ee96b104b1f9/scripts/sword_locales_to_json.mjs#L32 because this script is taking available locale list from this Ezra Bible App repository, it would be not a good practice to make the loop-back dependency from Node Sword Interface. So maybe this script should be back in this repository.
> 
> I'm putting all connected PR in draft until we will come to conclusion on this here

Here are some questions that could help us solve this riddle:

- Does the script make sense without the input from Ezra Bible App?
- Is it useful for node-sword-interface users without that language list from Ezra Bible App?
- Would it make sense to have part of it in node-sword-interface (the parsing part?) and part of it in ezra-bible-app? (using it as a library?)

In any case we should avoid the circular dependency.

zhuiks:
> * Move the new scripts underneith `build_scripts`.
> * Turn the action `convert-sword-locales` into a `postinstall` action in `package.json`.
> * Declare languages.json in git ignore

I think it should be a separate task/action because it's platform independent and the generated file wouldn't change that much. But the script requires node 14 to run

tobias-klein:
> > * Move the new scripts underneith `build_scripts`.
> > * Turn the action `convert-sword-locales` into a `postinstall` action in `package.json`.
> > * Declare languages.json in git ignore
> 
> I think it should be a separate task/action because it's platform independent and the generated file wouldn't change that much. But the script requires node 14 to run

Ok, that makes sense. Would you mind putting a note in the DESIGN.md document or alternatively in the CONTRIBUTING.md document.

zhuiks:
@tobias-klein All done here, this can be merged

tobias-klein (approved):

tobias-klein:
Thank you for your attention to details! Excellent technical improvement!

PR #412, opened 2021-08-13
Notes export

zhuiks:
Solves #96 

I decided to connect notes export to the same button that export verses with tags. That button was only active in "tags list" view.

Currently, it's implemented to export the whole chapter and user's notes, if any of the user's notes are in this chapter.

TODO:
- [ ] Convert markdown formatting to docx formatting
- [ ] Dispaly markdown list as docx lines
- [ ] Enable export on restoring from cache if any notes are present
- [ ] Export only one chapter in case if the whole book is loaded in the Bible browser
- [ ] Add translation name to docx
- [ ] Display translation copyright note


tobias-klein:
Just tried, but the export button does not seem to be active yet when a book is opened?

zhuiks:
> Just tried, but the export button does not seem to be active yet when a book is opened?

It's only works when you are opening book in chapter mode for now. And notes should be present in that chapter

tobias-klein:
> It's only works when you are opening book in chapter mode for now. And notes should be present in that chapter

What speaks against exporting the whole book?

zhuiks:
> > It's only works when you are opening book in chapter mode for now. And notes should be present in that chapter
> 
> What speaks against exporting the whole book?

There are different approaches on what exactly to export along with notes:
1. only export verse text that has user's notes
2. export all the verses of the bigger part of the Bible text (chapter) if any of the verses have notes
3. export all the verses of the whole Bible book if any of the verses have notes

I personally prefer the second approach, and that's how I implemented it. I just need to add some logic for the "whole book" browsing mode

tobias-klein:
> I personally prefer the second approach, and that's how I implemented it. I just need to add some logic for the "whole book" browsing mode

Ok! Clear :) Would be nice if it can be scaled / parameterized for the whole book mode.

zhuiks:
I switched to docxjs. It's a bit different. The main thing that left is to create/update nice formatting inside generated docx.

@tobias-klein let me know what do you think. I checked the generated document in LibreOffice, Word (online) and Google Docs. Everything seems to be the same, except fonts.

Also do you have an idea how to write a test for this feature. I was thinking about unit test, but the only thing that is easy to test is the file creation. Although, docx is just a zip file with a bunch of xmls. So we can go deeper.

tobias-klein:
> @tobias-klein let me know what do you think. I checked the generated document in LibreOffice, Word (online) and Google Docs. Everything seems to be the same, except fonts.

Great results so far! Well done! I have tested the export both in chapter mode and full book mode and it works nicely!
I am currently still getting some html entitities like `&quot;` and `&#39;` and `&amp;`.

> Also do you have an idea how to write a test for this feature. I was thinking about unit test, but the only thing that is easy to test is the file creation. Although, docx is just a zip file with a bunch of xmls. So we can go deeper.

Testing the content of the file would be more like an acceptance / integration test and it would be a lot of work I think. An alternative could be to first generate some kind of model for the document and then write the test only based on that model. But this would of course not be an end-to-end test.

**Other feedback**

I believe we need to show a disclaimer dialog before the export that shows the copyright notes from the respective translation. We can get the respective information by looking at the following attributes of the [module objects](https://github.com/ezra-bible-app/node-sword-interface/blob/master/API.md#moduleobject--object):

* distributionLicense
* shortCopyright

Maybe we can add a helper function in `sword_module_helper` that checks the content of the `distributionLicense` field and then returns true if a copyright exists? Then based on this we could show the content of the `shortCopyright` field in the dialog.

app/frontend/components/notes_tagged_verse_export.js
```diff
@@ -0,0 +1,128 @@
...
+   If not, see <http://www.gnu.org/licenses/>. */
+
+const exportController = require('../controllers/export_controller.js');
+
+/**
+ * The TaggedVerseExport component implements the export of tagged verses into a Word document.
+ * 
+ * @category Component
+ */
+class NotesTaggedVerseExport {
```
tobias-klein, line 26: To make it simply, maybe we should just name it DocExport?

app/frontend/components/notes_tagged_verse_export.js
```diff
@@ -0,0 +1,128 @@
...
+    currentTagIdList,
+    (verses, bibleBooks, groupedVerseTags) => { 
+      exportController.saveWordDocument(title, verses, bibleBooks); 
+    },
+    'docx',
+    false
+  );
+}
+
+function renderCurrentChapterNotesForExport(currentTab, isWholeBook=false) {
```
tobias-klein, line 105: So, since this function can also export the whole book, we should rename it to `renderNotesForExport`

app/frontend/controllers/export_controller.js
```diff
@@ -0,0 +1,579 @@
...
+ *
+ * @category Controller
+ */
+
+
+var exportFilePath;
+var docx;
+
+module.exports.showSaveDialog = async function (fileTitle) {
+  const dialog = require('electron').remote.dialog;
```
tobias-klein, line 34: Since we do depend on Electron APIs for the export, we should disable the functionality completely on Cordova/Android until we have found an alternative there (I didn't in my previous analysis).

tobias-klein:
I am getting the following error when opening the export of a tagged verse list. Will test with `master` to check whether this is a regression.

image

app/frontend/controllers/export_controller.js
```diff
@@ -0,0 +1,579 @@
...
+
+  // // Officegen calling this function after finishing to generate the docx document:
+  // docx.on('finalize', (written) => {
+  //   shell.openPath(exportFilePath);
+  // });
+
+  // // Officegen calling this function to report errors:
+  // docx.on('error', function(err) {
+  //   console.log(err);
+  // });
```
tobias-klein, line 64: To be removed.

app/frontend/controllers/export_controller.js
```diff
@@ -0,0 +1,579 @@
...
+  const shell = require('electron').shell;
+  shell.openPath(exportFilePath);
+  // var out = fs.createWriteStream(exportFilePath);
+
+  // out.on('error', function(err) {
+  //   console.log(err);
+  // });
+
+  // Async call to generate the output file:
+  // docx.generate(out);
```
tobias-klein, line 133: To be removed.

app/frontend/controllers/text_controller.js
```diff
@@ -713,16 +755,33 @@ class TextController {
...
-      await app_controller.initApplicationForVerseList(tabIndex);      
+      await app_controller.initApplicationForVerseList(tabIndex);
       uiHelper.hideTextLoadingIndicator();
     }
   }
 }
 
 module.exports = TextController;
 
+function filterNotesReferenceIds(verseNotes, startVerseNr, endVerseNr) {
```
tobias-klein, line 771: Wouldn't it make more sense to call it `getReferenceIdsFromNotes`?
The current name is a bit ambigous.

app/frontend/controllers/text_controller.js
```diff
@@ -713,16 +755,33 @@ class TextController {
...
+  const verseReferenceIds = Object.keys(verseNotes);
+  return verseReferenceIds.filter(verseReferenceId => {
+    const noteVerseNumber = parseInt(verseReferenceId.split('-')[2]);
+    return noteVerseNumber >= startVerseNr && (endVerseNr < startVerseNr || noteVerseNumber <= endVerseNr);
+  });
+}
+
+function filterNotesForVerses(verseNotes, startVerseNr, endVerseNr) {
```
tobias-klein, line 779: Wouldn't it make more sense to call it `getNotesForSection`. The function name does not indicate how we filter the notes.

package.json
```diff
@@ -81,7 +82,7 @@
     "marked": "^3.0.0",
     "mousetrap": "^1.6.5",
     "node-sword-interface": "^0.235.0",
-    "officegen": "^0.6.4",
+    "officegen": "^0.6.5",
```
tobias-klein, line 85: Remove this if we are not depending on it any longer.

zhuiks:
> I am getting the following error when opening the export of a tagged verse list. This is not happening on `master` with the same input data ... so it is a regression.
> 
> image

@tobias-klein 
What about notes export? Do you have the same error also?

app/frontend/controllers/export_controller.js
zhuiks: That's a good point. I'll try to research. I think it's a really important feature.

app/frontend/controllers/text_controller.js
zhuiks: Yes, thanks for pointing out. I think I started with one function and refactored it into two

app/frontend/controllers/text_controller.js
zhuiks: It doesn't "get notes" though. All notes come in `verseNotes` and returned only notes that are in range between `startVerseNr` and `endVerseNr`. I was thinking the name should be parallel to `Array.filter()`
Although, I guess you can call this "get".

tobias-klein:
I did not get the issue with the notes export. That worked flawlessly.

app/frontend/controllers/export_controller.js
tobias-klein: I think the key is whether you find a way to offer a file download from the context of the Android webview. Maybe there is a Cordova plugin for it? In the worst case I could look into creating one, but then I would have to brush up on my Java skills.

app/frontend/controllers/text_controller.js
tobias-klein: We can keep it this way, too!

app/frontend/components/notes_tagged_verse_export.js
zhuiks: done

app/frontend/controllers/export_controller.js
zhuiks: I'll disable it for now and leave a comment in the code

tobias-klein:
@zhuiks [TECH.md](https://github.com/ezra-bible-app/ezra-bible-app/blob/feature-export-notes/TECH.md) still needs to be updated, based on moving from officegen to docx.

tobias-klein:
Bildschirmfoto 2021-10-14 um 19 15 14

How about this example (tagged verse list export). Observations:

1) The verse references below the headlines are all upper case, but the headlines are not. Shouldn't we have it both in the same style?
2) It seems like foot note contents are also rendered (The additional numbers in the verse text). I think we should filter that out.

In the NET translation it looks like this in the source:

```html
<div class="sword-markup sword-note" osisref="Ps.16.7" n="15" style="display: inline-block; margin-left: 0.1em;">
<div class="sword-note-marker sword-marker" title="15">n</div>
</div>
```

Can we filter out all elements with class "sword-markup"?

tobias-klein:
> Theoretically every other than Public Domain license should cause the disclaimer dialog. However, the default KJV module comes with "General public license for distribution for any purpose". So this one is like a public license, right? What about Creative Commons?

Yes, even Creative Commons should be considered a license that we show to the user before running the export.

PR #415, opened 2021-08-14
Highlight strongs elements in currently selected verse on Cordova

tobias-klein:
Note that we could avoid the additional dependency between VerseSelection and DictionaryController in the future once the concept of the EventController would be available. In that case the verse selection would simply be another event that the DictionaryController could subscribe too.


zhuiks (approved):
No comments here whatsoever.
I wasn't able to test it on Android, but I trust that you did. Thank you!

PR #419, opened 2021-08-17
vscode launch.json fix

zhuiks:
The previous `chrome` type required an extra extension.
`pwa-chrome` type should work without anything extra

Please check by launching from debug view in VSCode. Confirm that break points are working.


tobias-klein:
> Please check by launching from debug view in VSCode. Confirm that break points are working.

Here is what I did:
1. Upgraded to latest VS Code on Linux (1.59.1)
2. Switch to this branch
3. Set a breakpoint
4. Launch the app
5. Try to trigger the functionality related to breakpoint

Result: Nothing happens! Does not seem to work ...
Note that I've never been using the VS Code Debugger before (I usually debug directly in the Electron Window).

tobias-klein:
I just tested this with the latest VSCode 1.63.0 and it is still not working.

zhuiks:
Looks like the `.vscode/launch.json` is already in master:
https://github.com/ezra-bible-app/ezra-bible-app/blob/master/.vscode/launch.json
I probably merged a commit from this to another branch. So no need to switch.

@tobias-klein When you start debug/run what configuration do you use? It should be "Debug Renderer Process". If you just debug the default node script, that would be the main Electron process.
Screenshot from 2021-12-12 21-22-25

tobias-klein:
> When you start debug/run what configuration do you use? It should be "Debug Renderer Process". If you just debug the default node script, that would be the main Electron process.

I've been using Debug Renderer Process. However, I am running into the bug that breakpoints are becoming de-attached right away.

I am going to close this PR for now. Maybe we can have another look in the future.

PR #423, opened 2021-08-22
New French translations

lemtom:
As always, I am not a native speaker so it would be best to find a native speaker to take over (and correct my translations).

One thing I didn't know how to account for: traduction and dictionnaire have different genders. So "est déjà installé" would change to "est déjà installée" depending on the gender, for example.


zhuiks:
> One thing I didn't know how to account for: traduction and dictionnaire have different genders. So "est déjà installé" would change to "est déjà installée" depending on the gender, for example.

Yes, I understand the problem. We got lucky with Slavic languages, as "dictionary" and "translation" is the same gender.

Let me check something and comeback with solution attempt

zhuiks:
@lemtom, I extended the code a bit, so it can process genders. The genders of "dictionary" and "translation" need to be specified in the `translation.json` as it is varies from language to language, and some languages don't even need that information:
```json
    "module-type-bible_singular": "traduction",
    "module-type-bible_gender": "masculine",
    "module-type-dict_singular": "dictionnaire",
    "module-type-dict_gender": "feminine",
```

Please pull this branch and correct, if I put wrong genders.

In the phrases that are gender specific, it can be used like this:
```json
      "module-already-installed_masculine": "{{module_type, singular-demonstrative-pronoun}} est déjà installé",
      "module-already-installed_feminine": "{{module_type, singular-demonstrative-pronoun}} est déjà installée",
```

Once again, please correct. I don't know French, I'm just guessing.
Let me know if this would work and if you have any questions.

PR #438, opened 2021-09-01
Upgrade electron to 13.2.3

### LINKED ISSUE #437: Upgrade to latest Electron 13.x

tobias-klein:
This should fix #437.


tobias-klein:
@zhuiks The tests are broken based on this Electron + Spectron upgrade.

This is where it fails:
features/steps/locale.js:34

```
const elementText = await element.getAttribute("innerText")
```

This does not work for some reason. I also tried `getText()` and `getAttribute('textcontent')`, but those are not working either. Maybe you can have a look?

zhuiks:
I think, newer version of Spectron is using newer version of WebDriver. So `getAttribute` will return only HTML element attributes.
From the very beginning, the proper should have been `getProperty`
`getText` is not working for some elements because according to [docs](https://webdriver.io/docs/api/element/getText) the element should be [interactable](https://www.w3.org/TR/webdriver/#interactability) which is not true for some elements that are checked.

tobias-klein:
Thank you so much @zhuiks!

PR #442, opened 2021-09-13
Optimize formatting of book selection menu by using four columns for OT

tobias-klein:
This looks better on a 7" tablet in portrait mode.

**Before**

Screenshot_20210913-104309

Screenshot (Sep 13, 2021 10_48_44)

**After**

Screenshot_20210913-103858

Screenshot_20210913-103851


css/main.css
```diff
@@ -621,17 +621,17 @@ main.content {
 
 .book-list-column {
   float: left;
-  width: 10.25em;
+  width: 8.4em;
```
zhuiks, line 624: This can be tricky with different languages, i.g Russian
Знімок екрану з 2021-09-16 07-31-38

css/main.css
```diff
@@ -621,17 +621,17 @@ main.content {
...
+  padding: 0.5em;
   padding-bottom: 0;
   margin-top: 0;
   margin-bottom: 0;
 }
 
 .ot-column {
   border-bottom: 1px solid lightgrey;
-  height: 26.5em;
+  height: 24em;
```
zhuiks, line 634: I'm getting an extra white space in portrait
Знімок екрану з 2021-09-16 07-34-56

zhuiks (changes requested):
In general, 4 columns for OT and 3 columns for NT make sense. 
But why are you splitting after "Ruth"?
I know for OT 5-12-5-5-12 is the nice division by genre.
How about if we try just 4x10 grid for OT and 3x9 grid for NT?

Let me add few commits...

tobias-klein:
Thanks for the feedback! We can divide differently based on well known schemes as you suggest!
Feel free to take it up ... I'm not going to get back to this in the next days.

zhuiks:
Here is what I got:
Знімок екрану з 2021-09-16 11-10-16
Знімок екрану з 2021-09-16 11-09-38

I decided to `text-align: center` because visually it's easier to identify a book by shape (name length).

zhuiks:
There is also a little bug when resizing from portrait to landscape with closed menu:
Знімок екрану з 2021-09-16 11-25-28

Seems like book menu dimensions are applied in JavaScript. I can make that with css

tobias-klein:
Thank you @zhuiks!
I am ok with the center alignment.

Additional thoughts:
1) I think it would be visually nicer if all the columns would be the same width, especially in portrait mode.
2) In portrait mode it is also visually nicer if the NT columns would align with the OT columns above. This was the case previously.

Regarding the bug you find. This is triggered within handleBookMenuClick when calling [`uiHelper.showButtonMenu`](https://github.com/ezra-bible-app/ezra-bible-app/blob/master/app/frontend/components/book_selection_menu.js#L232). That function explicitly sets the display style attribute to `block` when showing the menu - and that triggers the bug. Any suggestion how to address that?

zhuiks:
>  1. I think it would be visually nicer if all the columns would be the same width, especially in portrait mode.
> 
>  2. In portrait mode it is also visually nicer if the NT columns would align with the OT columns above. This was the case previously.

Is that what you meant (already committed):

Знімок екрану з 2021-09-24 15-19-24

zhuiks:
> Regarding the bug you found. This is triggered within handleBookMenuClick when calling [`uiHelper.showButtonMenu`](https://github.com/ezra-bible-app/ezra-bible-app/blob/master/app/frontend/components/book_selection_menu.js#L232). That function explicitly sets the `display` attribute to `block` when showing the menu - and that triggers the bug. Any suggestion how to address that?

The bug was with setting `display: block` for `#book-selection-menu-book-list` when closing book menu after selecting a chapter. I solved it by adding a class.

tobias-klein:
@zhuiks Any final review comments before I merge this? If not, then please approve via the review function.

css/main.css
```diff
@@ -588,8 +588,8 @@ main.content {
 
 @media only screen and (min-width: 770px) {
   #book-selection-menu-chapter-list {
-    width: 50.5em;
-    min-height: 20.8em;
+    width: 50.6em;
```
zhuiks, line 591: @tobias-klein have you tried to achieve the same width for book and chapter selection?

zhuiks (commented):
Please note that in some languages, e.g. Russian, book names are longer.
Width of the single column is controlled by `.book-list-column li` L#657

css/main.css
tobias-klein: Yes, that's what I tried. I'll have another look to see how this works with different languages. In general, it's difficult to replicate the width of the book selection to the chapter selection, because both elements use completely different layout methods ...

tobias-klein: I tested this now with multiple languages and currently it is not as it should be. The widths vary based on the languages used. And you cannot really calculate the overall width of the menu as before (I think before the column widths were static).

Another way would be maybe to not hide the book menu completely (`display: none`) when showing the chapter selection, but rather using `visibility: hidden`  and overlaying the chapter menu somehow using relative positioning.

Any additional suggestions?

zhuiks: How about if I move the width setting to the `#book-selection-menu` container?

zhuiks (changes requested):
@tobias-klein I think the right way would be to set up the desired width on the main `#book-selection-menu` container and make columns and children divs adjust.

css/main.css
```diff
@@ -619,48 +630,63 @@ main.content {
...
-    height: unset;
+  #book-selection-menu-book-list {
+    flex-direction: row;
   }
-  .book-list-column {
-    width: unset;
+
+  .nt-books {
+    grid-template-columns: repeat(3, 1fr);
+    flex-basis: 45%; /* 3 columns of 7 column layout + divider */
```
zhuiks, line 679: adjust this to make the columns equal width

zhuiks (commented):
Please, feel free to adjust the width

css/main.css
tobias-klein: Thanks!

tobias-klein:
I made some more adjustments and I still need to test this on a 7" tablet ...

zhuiks:
> I made some more adjustments and I still need to test this on a 7" tablet ... Meanwhile I feel like it would be best to have settings for two types of languages (compact vs. large) ... but optimization has to end somewhere, huh?

I think, there should be some middle point. For now I think it's 
I have some ideas how to reorganize book/chapter navigation. I'll share it under #299.

PR #449, opened 2021-09-20
Highlight frequently used tags in bold

### LINKED ISSUE #208: Book text: Render tags in bold if they are used multiple times in the same book
To get a better visual differentiation of tags that are used a couple of times vs tags that are only used a few times there could be a differentiation in formatting. If a tag is used a certain number of times in the current book then it could be rendered in bold. The number of occurances could either be relative based on the number of chapters or absolute (like at least three or four tagged verses in the current book).

Here is a quick prototype:
image

tobias-klein:
This resolves #208.

### DIFF
--- a/app/frontend/components/tags/tag_statistics.js
+++ b/app/frontend/components/tags/tag_statistics.js
@@ -17,33 +17,72 @@
    If not, see <http://www.gnu.org/licenses/>. */
 
 class TagStatistics {
-  constructor() {}
-
-  getBookTagStatistics() {
-    var global_tags_box_el = $('#tags-content-global');
-    var checkbox_tags = global_tags_box_el.find('.checkbox-tag');
-    var book_tag_statistics = [];
-
-    for (var i = 0; i < checkbox_tags.length; i++) {
-      var current_checkbox_tag = $(checkbox_tags[i]);
-      var current_checkbox_title = current_checkbox_tag.find('.cb-label').text();
-      var current_book_assignment_count = parseInt(current_checkbox_tag.attr('book-assignment-count'));
-      book_tag_statistics[current_checkbox_title] = current_book_assignment_count;
-    }
-
-    return book_tag_statistics;
+  constructor() {
+    this._frequentTagsList = [];
   }
 
-  async updateBookTagStatisticsBox(book_tag_statistics=undefined) {
-    if (book_tag_statistics === undefined) {
-      book_tag_statistics = this.getBookTagStatistics();
+  getBookTagStatistics() {
+    var globalTagsBoxEl = $('#tags-content-global');
+    var checkboxTags = globalTagsBoxEl.find('.checkbox-tag');
+    var bookTagStatistics = [];
+
+    for (var i = 0; i < checkboxTags.length; i++) {
+      var currentCheckboxTag = $(checkboxTags[i]);
+      var currentCheckboxTitle = currentCheckboxTag.find('.cb-label').text();
+      var currentBookAssignmentCount = parseInt(currentCheckboxTag.attr('book-assignment-count'));
+      bookTagStatistics[currentCheckboxTitle] = currentBookAssignmentCount;
     }
 
-    var tags_by_verse_count = Object.keys(book_tag_statistics).sort(
+    return bookTagStatistics;
+  }
+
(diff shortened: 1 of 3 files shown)


tobias-klein:
@zhuiks 

* Should we make this configurable?
* Would you prefer some other form of highlighting than bold text? (Like a color or underlining)
* Any other feedback?

zhuiks:
> @zhuiks
> 
>     * Should we make this configurable?
> 
>     * Would you prefer some other form of highlighting than bold text? (Like a color or underlining)
> 
>     * Any other feedback?


The first thing that comes to mind: maybe we can show a badge/chip with number of tagged verses?
Знімок екрану з 2021-09-20 19-05-55

Give me sometime to try to see it, I'll come with more concrete feedback

app/frontend/components/tags/tag_statistics.js
```diff
@@ -58,80 +97,79 @@ class TagStatistics {
...
+      let taggedVerseCount = this._currentBookTagStatistics[tag_title];
+      if (taggedVerseCount == 0) {
         continue;
       }
 
-      let tagged_verse_percent = Math.round((tagged_verse_count / overall_verse_count) * 100);
-      let current_row_html = "";
-
-      let isLessFrequent = tagged_verse_percent < MIN_CLUSTER_PERCENT || !clusters.includes(tagged_verse_percent);
+      let taggedVersePercent = Math.round((taggedVerseCount / overallVerseCount) * 100);
```
zhuiks, line 127: I think it would be better to have a percentage from the overall amount of tags in the book (or even in all the books).
If I have only 10-20 different tags, I want the tag with 3 occurrences to be highlighted.

zhuiks (changes requested):
Sorry being late to review this.

I think the tags highlighting should be independent of the book size.

app/frontend/components/tags/tag_statistics.js
tobias-klein: You mean from the overall amount of (used) tags or overall amount of tagged verses / tag assignments?

zhuiks: Overal amount of used tags. I think the idea that the number of occurances would be prety stable through out the different books and would vary only from overal tag amoun. I.e. even if I have just a few tags, most common tags should still be highlighted.

Everything else is ok.

tobias-klein: I'll implement this changed variant (changing the divisor in the formula) to see how it behaves.
Since tags / topics may vary greatly across books I would still suggest to keep it book-specific.

overallTaggedVerseCount should be easy to retrieve via [`ipcDb.getBookVerseTags`](https://github.com/ezra-bible-app/ezra-bible-app/blob/master/app/frontend/ipc/ipc_db.js#L99). I could use that instead of overallVerseCount. However, I have to filter out double verses.
But I think that's not what you indicate in your answer. What you mean is the number of tag assignments (one tag vs all tags), right?

How exactly would you define the "overall amount of used tags"?
This is easy to misinterpret ... (requirements :D)

zhuiks: I think maybe just use `taggedVerseCount * 0.1` for 10%. I think it should be simple and transparent for the end-user.
My initial thought was to have just a number of tag occurrences > NUMBER_OF_FREQUENT_TAGS. But then the problem is that most of your tags would be highlighted as you have a lot of them. But I think this is the simplest option for the end user to understand why it's highlighted (in bold).

tobias-klein: The way how I adapted it now is this:

```
  getTagPercentage(bookTagStatistics, tagTitle, overallTagCount) {
    let taggedVerseCount = bookTagStatistics[tagTitle];
    let taggedVersePercent = Math.round((taggedVerseCount / overallTagCount) * 100);
    return taggedVersePercent;
  }
```

This in combination with `MIN_CLUSTER_PERCENT` (2) and `MAX_CLUSTERS` (4) then determines, which tags are most frequently used. It works well both with small collections of tags and big collections of tags.

Regarding the highlighting I think eventually we need to integrate some online help / hints into the app to help the user understand.

For now, I would like to keep it like this. Of course we can keep iterating based on personal usage experience and additional ideas!

PR #464, opened 2021-10-08
Add Brazilian Portuguese locale

tobias-klein:


zhuiks:
> @zhuiks Note that the following code in I18nController causes a bug related to this new locale, because the locale uses the four letter code pt-BR. The bug I have observed is that the current language (item selected in language selection box) is not correctly reflected based on the locale code pt-BR.
> 
> Where do we need the two letter code specifically?
> 
> ```js
> /** returns current app locale (2-letter language code) */
> module.exports.getLocale = function() {
>   var locale = i18n.language;
>   return locale.slice(0, 2); // just in case we got language with the region code (i.e "en-US") we want only the language code ("en")
> }
> ```

I think there was an issue with switching/detecting the right language, specifically for English: en-US and en-GB. But this code snippet is not a proper solution.

zhuiks:
1. When we add new language, we need to regenerate language data with `npm run convert-sword-languages` (this is in CONTRIBUTING.md p.6).
2. Locale data in generated [`languages.json`](https://github.com/ezra-project/ezra-project/blob/080b3b73688214dc883876d81747ede6f523c510/lib/languages.json) would match available locale codes, including their case (upper case for regions), so no need for lower case.
3. I removed `getFullLocale` and made `getLocale` return full locale code (including region).
4. This reintroduced bug #467, which can be resolved under another PR.

@tobias-klein let me know if this is not clear, and you still have questions. This is probably easier to discuss over skype.

PR #500, opened 2021-10-30
Hide tag assignment menu after assignment changes

### LINKED ISSUE #445: Close tag assignment menu immediately after assign/unassign action

tobias-klein:
This will fix #445.

### DIFF
--- a/app/frontend/components/tags/tag_assignment_menu.js
+++ b/app/frontend/components/tags/tag_assignment_menu.js
@@ -57,7 +57,7 @@ class TagAssignmentMenu {
   hideTagAssignmentMenuAfterDelay() {
     setTimeout(() => {
       this.hideTagAssignmentMenu();
-    }, 500);
+    }, 600);
   }
 
   async handleMenuClick(event) {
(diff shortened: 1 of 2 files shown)


tobias-klein:
@zhuiks 
This is a rather simple change. The tag assignment menu is hidden after a 500ms delay. Can you check whether this delay feels right from a user perspective? Should we increase it slightly or even make it shorter? If the menu is hidden without a delay then that is too sudden after the assignment/unassignment of a tag. That's why I added it.

app/frontend/components/tags/tag_assignment_menu.js
```diff
@@ -54,6 +54,12 @@ class TagAssignmentMenu {
     }
   }
 
+  hideTagAssignmentMenuAfterDelay() {
+    setTimeout(() => {
+      this.hideTagAssignmentMenu();
+    }, 500);
```
zhuiks, line 60: @tobias-klein I think we can slightly increase it to 600ms

zhuiks (changes requested):
We can slightly increase the delay.

app/frontend/components/tags/tag_assignment_menu.js
tobias-klein: Ok!

Issue #282, opened 2021-05-16
Refactoring: Use observer pattern for events currently handled by AppController


tobias-klein:
We could use the observer pattern for handling these events:

* tab search results available => currently `AppController.onTabSearchResultsAvailable`
* tab search reset => currently `AppController.onTabSearchReset`
* tab selected => currently `AppController.onTabSelected`
* tab added => currently `AppController.onTabAdded`
* bible translation changed => currently `AppController.onBibleTranslationChanged`
* all translations removed => currently `AppController.onAllTranslationsRemoved`
* translation removed => currently `AppController.onTranslationRemoved`

This currently corresponds to ~200 lines of code in `AppController` ... which would then mostly move to the respective components that respond to these events.

tobias-klein:
@zhuiks Any thoughts on this?

zhuiks:
Big yes for refactoring! AppController and TabController were really hard to figure out what runs after what and what runs only at start up.

However, I think refactor Tab to tab web component should come first. It can be 3-4 components based on tab type with common base tab component. 
Also the decision needs to be made about state management. Having clearly defined state object/objects helps to understand component dependencies on the state parts. Observer pattern by itself doesn't provide that clarity.

zhuiks:
My thoughts:
 
The new `eventController` will have the implementation of the observer pattern.

Each component that generates the event should create with the help of `eventController` and export an appropriate subscription method (i.e. `onTranslationRemoved`) and notify all subscribers by triggering something like `eventController.trigger('TranslationRemoved')`

Each subscriber component would subscribe with that method:
```javascript
app_controller.moduleAssistatant.onTranslationRemoved(() => {
  // do something inside the subscriber component
});
```

That way we will have the central implementation. But at the same time, all components would be responsible for their events. And we would have a descriptive name for the subscription.

tobias-klein:
Good design ideas! I could imagine a design like that!
The design pattern is essentially like a central "message bus" for events.

PR #521, opened 2021-11-08
Implement EventController

zhuiks:
This solves #282

Observer is implemented as pub/sub pattern having `eventController` as event bus.

Pending events for porting: 

- [x] on-start-repo-update
- [x] on-progress-repo-update
- [x] on-complete-repo-update
- [x] on-bible-text-loaded
- [x] on-translation-added
- [x] on-translation-removed
- [x] on-all-translations-removed
- [x] on-translation-changed
- [x] on-tab-selected
- [x] on-tab-added
- [x] on-tab-search-results-available
- [x] on-tab-search-reset
- [x] on-locale-changed
- [x] on-theme-changed
- [x] on-fullscreen-changed

Aside from this we should also:
- [x] document the event controller and the related pattern in DESIGN.md.

### DIFF
--- a/app/frontend/controllers/event_controller.js
+++ b/app/frontend/controllers/event_controller.js
@@ -36,7 +36,7 @@ function notCreated(event) {
  * @typedef { "on-start-repo-update" | "on-progress-repo-update" | "on-complete-repo-update" } RepoUpdateEvents
  */
 /**
- * @typedef { "on-bible-text-loaded" } BibleTextEvents
+ * @typedef { "on-bible-text-loaded" | "on-bible-translation-changed" } BibleTextEvents
  */
 /**
  * @typedef { "on-all-translations-removed" | "on-translation-removed" } ModuleAssistantEvents
--- a/app/frontend/controllers/event_controller.js
+++ b/app/frontend/controllers/event_controller.js
@@ -39,8 +39,11 @@ function notCreated(event) {
  * @typedef { "on-bible-text-loaded" } BibleTextEvents
  */
 /**
+ * @typedef { "on-all-translations-removed" | "on-translation-removed" } ModuleAssistantEvents
+ */
+/**
  * For consistency event name should be a string in-kebab-case with the following template: "on-[action]-[object]"
- * @typedef { RepoUpdateEvents | BibleTextEvents } Event
+ * @typedef { RepoUpdateEvents | BibleTextEvents | ModuleAssistantEvents } Event
  */
 
 /**
--- a/app/frontend/controllers/event_controller.js
+++ b/app/frontend/controllers/event_controller.js
@@ -1,0 +1,88 @@
+/* This file is part of Ezra Bible App.
+
+   Copyright (C) 2019 - 2021 Ezra Bible App Development Team <contact@ezrabibleapp.net>
+
+   Ezra Bible App is free software: you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation, either version 2 of the License, or
+   (at your option) any later version.
+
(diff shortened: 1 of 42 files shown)


app/frontend/controllers/event_controller.js
```diff
@@ -0,0 +1,136 @@
...
+  return !(event in messages);
+}
+
+/**
+ * @typedef { "on-start-repo-update" | "on-progress-repo-update" | "on-complete-repo-update" } RepoUpdateEvents
+ */
+/**
+ * For consistency event name should be a string in-kebab-case with the following template: "on-[action]-[object]"
+ * @typedef { RepoUpdateEvents } Event
+ */
```
tobias-klein, line 41: I would suggest to move these typedefs to the place where these events are published.

Otherwise we have to keep touching this controller whenever we add events.

app/frontend/controllers/event_controller.js
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * This controller implements pub/sub pattern and serves as a broker (event bus) between
+ * event emitter (publisher) and event consumer subscriber.
+ * https://en.wikipedia.org/wiki/Publish%E2%80%93subscribe_pattern
+ * 
+ * @module eventController
+ * @category Controller
+ */
+
+var messages = {};
```
tobias-klein, line 29: Why `messages` and not `subscribers`?

tobias-klein (commented):
Excellent! Simple and powerful. I like this implementation and just have a few comments. Thanks also for implementing the unit test!

tobias-klein:
@zhuiks I applied this to see how it works and transformed AppController.initApplicationForVerseList into the event 'on-bible-text-loaded' which is now triggered by TextController. It works nicely!

app/frontend/controllers/event_controller.js
zhuiks: My initial thought was that publisher delivers _messages_ to subscribers according to the common pattern.
But this variable hold subscribers and not messages. So I'll change that.

app/frontend/controllers/event_controller.js
tobias-klein: I get it now based on the auto completion I have seen while working on it. However, would it make sense to move all event typedefs to a dedicated file? (like events.js) Or would that then mean that all subscribers would need to require two modules instead of just one?

app/frontend/controllers/event_controller.js
```diff
@@ -0,0 +1,142 @@
...
+ * @typedef { "on-start-repo-update" | "on-progress-repo-update" | "on-complete-repo-update" } RepoUpdateEvents
+ */
+/**
+ * @typedef { "on-bible-text-loaded" } BibleTextEvents
+ */
+/**
+ * @typedef { "on-all-translations-removed" | "on-translation-removed" } ModuleAssistantEvents
+ */
+/**
+ * For consistency event name should be a string in-kebab-case with the following template: "on-[action]-[object]"
```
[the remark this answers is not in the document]
tobias-klein: Yes. And I agree that consistency is important. I think having the verb/action at the end is more natural. Thanks in advance for changing the other instances!

tobias-klein:
Seems like the last commits broke the Cucumber tests.

These are the [scenarios that fail](https://github.com/ezra-bible-app/ezra-bible-app/runs/4185738361?check_suite_focus=true):
1) Scenario: Basic search # features/005_module_search.feature:25
2) Scenario: Adding a note to a book # features/006_notes.feature:28

zhuiks:
@tobias-klein thanks for picking it up and fixing!

tobias-klein (approved):

tobias-klein:
Good team work on this one @zhuiks! Thank you so much!

PR #526, opened 2021-11-19
Side panel buttons

### LINKED ISSUE #346: Make sidebar (tag view) and dictionary switchable by side buttons
As discussed in #299 it would be nice to be able to switch toolbar/sidebar and dictionary without opening option dialog

zhuiks:
This closes #346


zhuiks:
@tobias-klein fill free to check it out.
Under this I also refactored some styles with css variables.

tobias-klein:
Cool, nice! :+1:

I'm wondering whether it would make sense to implement it like in VS Code. Having one overall sidebar like the "Explorer" in VS Code that allows you to hide the complete sidebar with one click. This is not possible with the current implementation because there we handle tag list and dictionary independently. If we would instead have a "side bar" and have the tag list and dictionary as children of that (similar to "Open Editors", "Workspace", "Outline", etc. in the VS Code "Explorer") we would be able to both toggle the child components within the side bar, but then also hide the complete side bar at once just with one click. Does that make sense? I think I would personally prefer that behavior, because I definitely have the usecase to hide the complete side bar at once, and I would like to do that with one click.

zhuiks:
> I'm wondering whether it would make sense to implement it like in VS Code. Having one overall sidebar like the "Explorer" in VS Code that allows you to hide the complete sidebar with one click. This is not possible with the current implementation because there we handle tag list and dictionary independently. 

Yes, that make sense. 
What I'm thinking though is that buttons at the bottom would switch on/off footer area (currently dictionary moves there in tablet/portrait mode). Instead of switching the tag assignment to the menu button for this mode, I'm planning to make it show over the Bible browser content. 
Having dictionary in the footer make sense as you want to see both dictionary articles and Bible text.

tobias-klein:
> Instead of switching the tag assignment to the menu button for this mode, I'm planning to make it show over the Bible browser content.

I am curious how you will solve this in portrait mode. Another usecase is the new fullscreen mode.

zhuiks:
>I am curious how you will solve this in portrait mode. And aside from that - another usecase is the new fullscreen mode.

That's the result with the latest commit (there is a glitch in the light mode):
Screenshot from 2021-11-20 21-03-50

When switching to full-screen, the side panel would be automatically hidden.

tobias-klein:
@zhuiks 
I like it! However, I have one particular concern in regard to usability. When assigning tags it would be nice to still see the verse text that you are assigning text to. With this kind of UI the verse text would be covered for the most part. What we could do is add a box within the verse list (maybe on top) where we just copy/visualize the currently selected verse(s), so that you still see the verse content while assigning tag(s). In any case I would suggest to implement that as a separate issue.

zhuiks:
I have another idea:
https://github.com/ezra-bible-app/ezra-bible-app/discussions/299#discussioncomment-1685073

zhuiks:
I think we can merge this, and I can start working on #542.

tobias-klein:
I just checked this out on my surface tablet and this is how it looked right after the startup.

tobias-klein:
Can we reduce the horizontal width of the sidebar a bit?
Here is the horizontal width compared with the sidebar in VS Code, which could serve as a reference.

image

tobias-klein:
In portrait mode, the tab bar is visible behind the tag list somehow. You see it with the additional blue line underneith the title area of the tag list.

image

tobias-klein:
Screenshot of portrait mode on Surface Go tablet.

image

I think I like the idea of it, but currently this is not usable for me in portrait mode - for the following reasons: Once the tag list is shown it occupies most of the screen space and the Bible text is hardly visible anymore. This essentially forces me to context switch between the Bible text and the tag list. Whenever I am assigning/unassigning tags I would still like to see the Bible text (especially the verse that I am working with right now) and not have the tag list cover up the complete Bible text.

Here's what we could do in portrait mode:
* Reduce the width of the side bar to something significantly smaller (20-30% less than what it is currently).
* Show the currently selected Bible verse(s) in an overlay somehow.
* In portrait mode the side bar does not necessarily have to occupy the whole vertical space.
* We could also say that we do not use the left sidebar at all in portrait mode and also put the tag list in the area where the dictionary is shown at this point in portrait mode.

tobias-klein:
Here are some more glitches I observed:
* It doesn't seem to save the settings that were there when entering fullscreen mode. Given I am in regular mode (not fullscreen) and the sidebar is hidden when I enter the fullscreen mode and leave it again then the side bar is popping up even though it was not visible before I entered the fullscreen mode.
* After changing the screen orientation of my tablet I somehow ended up with this formatting and then after this the vertical height of the side bar stayed like this:
image

zhuiks:
@tobias-klein Thanks for feedback. Based on it, it makes more sense to implement #542 under this PR. Some glitches then wouldn't be relevant, as we have to worry only about portrait or landscape. Probably I have to introduce a new setting: what tool panel is shown in place of tag list shown and dictionary shown.

I also propose to use the term tool panel (which can be either side panel or bottom panel based on orientation).
Then the particular window can be "tag panel" or "dictionary panel".

zhuiks:
Last commits close #542.

@tobias-klein what should we do with resize? Ideally, tool panel should resize horizontally for landscape and vertically for portrait.

Do we still need `tag_assigment` with the latest layout updates?

tobias-klein:
> Last commits close #542.

Excellent!

Here is a screenshot on macOS, but I also saw this effect on Windows. When switching from tag list to dictionary, for a split second the formatting / style is not yet fully applied. Can we fix that?

Bildschirmfoto 2021-12-02 um 18 37 19

tobias-klein:
> @tobias-klein what should we do with resize? Ideally, tool panel should resize horizontally for landscape and vertically for portrait.

Yes. I agree. And on the desktop this would be simply based on the usual click and drag motion using the mouse. Not sure how to do this on a touchscreen, though - maybe there we need to simply start with reasonable default dimensions and think about a function to adjust this in a second step?
 
> Do we still need `tag_assigment` with the latest layout updates?

What do you mean exactly? The button to assign the last used tag?

tobias-klein:
Looking at the portrait vertical height partitioning it seems that the tag list height is fixed whereas the verse list resizes dynamically. I would say it should be the other way around. The verse list should have a minimal height and not get smaller anymore while the tag list height could be more dynamic.

Bildschirmfoto 2021-12-02 um 18 45 15

tobias-klein:
I have a hard time recognizing which is the active button and which is the inactive one ...
Bildschirmfoto 2021-12-02 um 18 56 13

app/frontend/event_types.js
```diff
@@ -41,9 +41,12 @@
...
  */
+/**
+ * @typedef { "on-tag-panel-switched" | "on-dictionary-panel-switched" } PanelsEvents
+ */
 
 
 
 /**
- * @typedef { BibleTextEvents | TabEvents | TabSearchEvents | ModuleAssistantEvents | RepoUpdateEvents | I18nEvents | AppEvents } EzraEvent
+ * @typedef { BibleTextEvents | TabEvents | TabSearchEvents | ModuleAssistantEvents | RepoUpdateEvents | I18nEvents | AppEvents | PanelsEvents } EzraEvent
```
tobias-klein, line 51: How about `PanelEvents` instead of `PanelsEvents`.

app/frontend/helpers/ui_helper.js
```diff
@@ -121,52 +118,46 @@ class UiHelper {
...
       
-      if (!platform.isFullScreen() && app_controller.optionsMenu._tagListOption.isChecked) {
-        app_controller.tag_assignment_menu.moveTagAssignmentList("SIDE_PANEL");
-      }
+      // if (!platform.isFullScreen() && app_controller.optionsMenu._tagListOption.isChecked) {
+      //  app_controller.tag_assignment_menu.moveTagAssignmentList("SIDE_PANEL");
+      // }
 
-      app_controller.dictionary_controller.moveInfoBoxFromTo(bottomPanel, sidePanel);
+      // app_controller.dictionary_controller.moveInfoBoxFromTo(detailPanel, toolPanel);
```
tobias-klein, line 145: Let's remove this if we don't need it anymore.

features/steps/toolpanel.js
```diff
@@ -0,0 +1,31 @@
+/* This file is part of Ezra Bible App.
```
tobias-klein, line 1: To be consistent I would name the file `tool_panel.js` instead of `toolpanel.js`.

tobias-klein (commented):
Just some minor points concerning the code.

zhuiks:
> > Do we still need `tag_assigment` with the latest layout updates?
> 
> What do you mean exactly? The button to assign the last used tag?

So the whole [`tag_assignment_menu.js`](https://github.com/ezra-bible-app/ezra-bible-app/blob/05161b9fafe8063e0aa3d2cd2b78f0cffa57e35f/app/frontend/components/tags/tag_assignment_menu.js#L28) is not relevant anymore?
Also, for full-screen wheel nav can tag assignment option be deleted since it's available via tool panel button?

tobias-klein:
@zhuiks No, I don't think we need the tag_assignment_menu anymore, with the changes of this PR it is obsolete now.

Yes, for the fullscreen mode we should consequently also remove that option.

zhuiks:
Now it looks like final :)

I also cleaned up `uiHelper`. All resizing now is pure css! @tobias-klein let me know if I missed/broke something. 
Resize performance can be improved. But it can be done under new issue. It's not that critical.

tobias-klein:
Thank you so much @zhuiks!
This is a great engagement to the UI!
I've been adding a few more style optimizations for Android and there are still some small points left. I'll finish this up and then merge!

tobias-klein (dismissed):

tobias-klein (dismissed):

app/frontend/startup.js
```diff
@@ -312,7 +312,8 @@ class Startup {
     $('#loading-subtitle').hide();
 
     // Show main content
-    $('#main-content').show();
+    document.getElementById('main-content').style.display = 'block';
+    document.getElementById('panel-buttons').style.display = 'block';
```
zhuiks, line 316: @tobias-klein `panel-buttons` are inside `main-content`. Is there a problem displaying it?

tobias-klein: There was not a problem displaying them, but rather a problem that they were displayed too early. I noticed this when I tested it on Android. They were already visible at the point when the permissions message was shown at initial startup, but at that point nothing is supposed to be visible aside from that dialog with the permissions message. This adjustment (`display: none`) by default and explicitly showing it solved the issue. If we find a better solution later, we can still change this.

zhuiks: I guess my question was why should `panel-buttons` also be hidden and reshown as it is a descendant of `main-content` which the default hidden by default.

tobias-klein: Screenshot (Dec 5, 2021 10_15_39)

This is how it looked on Android before my latest change. You see that there are some panel buttons shown - based on the element in index.html (even though not initialized yet).

The reason is that the panel-buttons are (contrary to other components) in the DOM (index.html) right from the beginning and were not hidden by default. Looking at the startup sequence, other components are only loaded after the Android permissions have been handled.

What happens before the Android permissions box is shown is that this function is called:
```JavaScript
  hideGlobalLoadingIndicator() {
    var loadingIndicator = $('#startup-loading-indicator');
    loadingIndicator.hide();
    $('#main-content').show();
  }
```

At that point the panel-buttons were then becoming visible, even though they shouldn't. Other components didn't show the same issue, because they weren't even loaded in the DOM at that point. Note that after the Android permissions are acknowleged by the user, the global loading indicator is shown again (and main-content hidden), consider this piece of code from `cordova_platform.js`:

```JavaScript
  onPermissionGranted() {
    console.log("Permission to access storage has been GRANTED!");
    this.getPermissionBox().dialog('close');
    uiHelper.showGlobalLoadingIndicator();
    this.initPersistenceAndStart();
  }
```

As a fix I made the panel-buttons invisible by default (`display: none` in main.css) and am activating them now during the startup sequence. To make it easier to understand I moved the activation to the function `loadHTML` (in startup.js) with some comment.

tobias-klein (approved):

app/frontend/startup.js
zhuiks: @tobias-klein thank you for detailed explanation! Now I understand the problem.
I know you already merge this, so I have a bit different idea in another PR.

PR #554, opened 2021-12-05
Extract more functionality from AppController into separate controllers

tobias-klein:
This PR extracts two new Controllers from the AppController:

* **ReferenceVerseController**: Handles the lifecycle of the ReferenceVerse box on top of the verse list, used both in case of x-refs as wells tagged verse lists.
* **VerseListController**: Provides an API for the verse list as well as event handlers clicks within the verse list.


tobias-klein:
@zhuiks This contributes to a more lean AppController. I'd be glad if you could have a brief look in terms of code review. Note that the content of functions has not been modified very much. Functions were simply moved from AppController to new Controllers. Interaction with the verseList now happens via verseListController instead of AppController.

zhuiks:
I'll review it this weekend.

app/frontend/controllers/reference_verse_controller.js
```diff
@@ -0,0 +1,130 @@
...
+}
+
+module.exports = {
+  showReferenceContainer,
+  getCurrentReferenceVerse,
+  getLocalizedReferenceVerse,
+  updateReferenceVerseTranslation,
+  clearReferenceVerse,
+  renderReferenceVerse
+};
```
zhuiks, line 130: I think, for new modules, there should be consistency with exports. Previously we used 
```javascript
module.exports.renderReferenceVerse = async function (verseBox, tabIndex=undefined) {
```
Instead, we can switch to this rewrite of `module.eports` object (Although some people say that it might be a problem in rear cases, which I don't think we should worry about). It's just should be consistent throughout new/refactored code.

Personally, I like what we used, as it shows which functions/methods are external right away.

app/frontend/controllers/verse_list_controller.js
```diff
@@ -0,0 +1,364 @@
...
+  showVerseListLoadingIndicator,
+  hideVerseListLoadingIndicator,
+  getBibleBookStatsFromVerseList,
+  getFirstVisibleVerseAnchor,
+  resetVerseListView,
+  getVerseListBookNumber,
+  bindEventsAfterBibleTextLoaded,
+  bindXrefEvents,
+  handleReferenceClick
+};
```
zhuiks, line 364: Same comment as with `referenceVerseController`. There should be consistency with new/refactored code regarding `module.exports`.

zhuiks (approved):
Nice!
I didn't notice anything that I don't like. I like having thin `app_controller` and minimized `uiHelper`.
`reference_separator` should not be global. But we have an issue about that.

The only comment that I have is about consistency style of `module.eports` for new/refactored code.

app/frontend/controllers/reference_verse_controller.js
tobias-klein: Thanks! Consistency is important - I agree. I will change to the `module.exports.functionname` ... style.

Issue #555, opened 2021-12-06
Unstyled panel buttons showing up with permission dialog.


zhuiks:
![Screenshot (Dec 5, 2021 10_15_39)](https://user-images.githubusercontent.com/17739895/144741218-2832e742-2604-46ca-ad39-13d929f01226.png)

This is how it looked on Android before my latest change. You see that there are some panel buttons shown - based on the element in index.html (even though not initialized yet).

The reason is that the panel-buttons are (contrary to other components) in the DOM (index.html) right from the beginning and were not hidden by default. Looking at the startup sequence, other components are only loaded after the Android permissions have been handled.

What happens before the Android permissions box is shown is that this function is called:
```JavaScript
  hideGlobalLoadingIndicator() {
    var loadingIndicator = $('#startup-loading-indicator');
    loadingIndicator.hide();
    $('#main-content').show();
  }
```

At that point the panel-buttons were then becoming visible, even though they shouldn't. Other components didn't show the same issue, because they weren't even loaded in the DOM at that point. Note that after the Android permissions are acknowleged by the user, the global loading indicator is shown again (and main-content hidden), consider this piece of code from `cordova_platform.js`:

```JavaScript
  onPermissionGranted() {
    console.log("Permission to access storage has been GRANTED!");
    this.getPermissionBox().dialog('close');
    uiHelper.showGlobalLoadingIndicator();
    this.initPersistenceAndStart();
  }
```

As a fix I made the panel-buttons invisible by default (`display: none` in main.css) and am activating them now during the startup sequence. To make it easier to understand I moved the activation to the function `loadHTML` (in startup.js) with some comment.

_Originally posted by @tobias-klein in https://github.com/ezra-bible-app/ezra-bible-app/pull/526#discussion_r762537243_

PR #556, opened 2021-12-06
Refactor panel button optimizations

zhuiks:
This solves #555


zhuiks:
@tobias-klein I think this way is a better solution for #555

tobias-klein:
Thanks @zhuiks! I'll test it!

tobias-klein (approved):
Thank you @zhuiks!
I just tested this and it works!

Issue #572, opened 2021-12-19
Move New Tag button to tag list panel


tobias-klein:
The New Tag button should be moved to the tag list panel, since this function is so closely related to that panel.

Prototype:

![image](https://user-images.githubusercontent.com/17739895/146680363-2aceffc3-3c82-401f-b709-811bfa2d07a2.png)

tobias-klein:
@zhuiks @augustin-colesnic @debritto
This is a prototype. I put the "new tag button" just with the icon (no text) right next to the filter field on top of the tag panel.

Does this make sense from your perspective?

debritto:
Hi @tobias-klein !
Yes, it makes sense. 
This strategy will guarantee a clean and sleek design of the application. 
Great idea!

PR #582, opened 2021-12-20
Various layout improvements

### LINKED ISSUE #581: Move compare translation function to tool panel
Acceptance criteria:

* Move the compare translation function to the toolbox panel.
* Remove button menu.
* Update automatically on verse selection.
* Make first column with translations smaller (use translation ids).

tobias-klein:
This fixes #581 and #594 and #572 and #595.

* Utilize side panel for compare translations function and tag statistics
* Move new tag button to tag list
* Move verse context buttons to separate context menu to save space

Current draft implementation (outdated):
compare_translations_side_panel


zhuiks:
Wow @tobias-klein, you've been busy :) Great job!!!
Merry Christmas!

I like having #581 "Compare translations" in the tool window.

For #572 I propose to move filter, search, and new tag button to a new row. It would add a more white-space/breathing room, so visually it doesn't feel so crammed:
Tag list

#594 Tag statistic button in the toolbar can be disabled if there are no tags, otherwise it's very confusing:
Screenshot from 2021-12-27 10-10-14

With #595 menu for selected verses: 
I think, eventually, it would be better to have all the actions as an overlay above/below selected verse(s). But for now that works. Icon can be different. Maybe the button can be joined and have a text "Selected verses: Phil 1:21".

I'll review the code later.

tobias-klein:
Merry Christmas @zhuiks :). Thanks for your feedback!

> For #572 I propose to move filter, search, and new tag button to a new row. It would add a more white-space/breathing room, so visually it doesn't feel so crammed:

Yes, that makes sense.

> #594 Tag statistic button in the toolbar can be disabled if there are no tags, otherwise it's very confusing.

I agree! The alternative would be to show a help message like for the dictionary panel and comparison panel.

> With #595 menu for selected verses:
> I think, eventually, it would be better to have all the actions as an overlay above/below selected verse(s). But for now that works. Icon can be different. Maybe the button can be joined and have a text "Selected verses: Phil 1:21".

Good idea! I'll look into this option.

tobias-klein:
@zhuiks And btw ... It was fun working with the new panel buttons component for extensions :). Very easy to work with from a coding perspective.

tobias-klein:
Btw ... this is the usecase for the minimum screen size.

7" tablet with active onscreen keyboard.

image

app/frontend/controllers/app_controller.js
```diff
@@ -173,10 +176,10 @@ class AppController {
...
   }
 
-  initVerseButtons(tabIndex=undefined) {
-    var currentVerseListMenu = app_controller.getCurrentVerseListMenu(tabIndex)[0];
-    var editNoteButton = currentVerseListMenu.querySelector('.edit-note-button');
-    var copyButton = currentVerseListMenu.querySelector('.copy-clipboard-button');
+  initVerseContextButtons() {
+    var verseContextMenu = document.getElementById('verse-context-menu');
+    var editNoteButton = verseContextMenu.querySelector('.edit-note-button');
+    var copyButton = verseContextMenu.querySelector('.copy-clipboard-button');
```
zhuiks, line 182: How about moving verseContextMenu into a separate component along with `initVerseContextMenu()`, `toggleVerseContextMenuButton()`, `enableVerseButtons()`, `disableVerseButtons()` and `hideVerseContextMenu()` methods?

css/main.css
```diff
@@ -458,10 +458,16 @@ h2:nth-of-type(1) {
...
-#display-options-menu {
+#display-options-menu,
+#verse-context-menu {
   font-size: 0.9em;
 }
 
+#verse-context-menu .fg-button {
+  margin: 0.8em;
+  padding: 0.5em;
+}
```
zhuiks, line 469: All `#verse-context-menu` styles can be inside verseContextMenu component

html/verse_context_menu.html
```diff
@@ -0,0 +1,18 @@
...
+
+<div class="copy-clipboard-button fg-button ui-state-default ui-corner-all ui-state-disabled">
+  <i class="fas fa-clipboard"></i>
+  <span i18n="bible-browser.copy"></span>
+</div>
+
+<div class="show-context-button fg-button ui-state-default ui-corner-all ui-state-disabled">
+  <i class="fas fa-arrows-alt-v"></i><i class="fas fa-align-justify"></i>
+  <span i18n="tags-toolbar.context"></span>
+</div>
```
zhuiks, line 18: This can be a web-component. 
It would be much easier to extend/refractor it if all view-layout, css style and js logic would be in one place (file).

zhuiks (changes requested):
I guess the only request I have is to implement verseContextMenu as a web-component. It will be less coupled with other parts that way.

There is always more to refactor, like remove dependencies on jQuery. But you've done a huge work under this PR, @tobias-klein! And we are not at the point of saying goodbye to jQuery any time soon. So I didn't request any changes about that.

Thank you for moving translation strings inside translation files. That will be much easier for translators to deal with.

zhuiks:
@tobias-klein I adjusted tag edit/remove button styles. I also converted assign tag checkbox to a button.
Let me know if this is something you want to keep or adjust.

tobias-klein:
@zhuiks Thank you!

> I adjusted tag edit/remove button styles. I also converted assign tag checkbox to a button.
> Let me know if this is something you want to keep or adjust.

I like the mouseover hover styles of the buttons.
Regarding the tag symbol instead of the checkbox I have to get used to this. :)

But independent of what I think ... I'd be interested what other users say.
@debritto @augustin-colesnic
How do you like the tag buttons instead of checkboxes? Consider the screenshot below.

image

app/frontend/controllers/app_controller.js
tobias-klein: Yes, that makes sense :+1:

debritto:
Hi @tobias-klein !
I personally like the idea of a clickable tag button instead of a checkbox.
Can 't wait to see new version in action ;)

app/frontend/controllers/app_controller.js
tobias-klein: I moved the verse context menu into a separate component now. I kept it without shadow-dom, because that was the easiest. And the menu button is still separate. It's not perfect this way, but I think sufficient for now. Feel free to have another look.

tobias-klein:
> For #572 I propose to move filter, search, and new tag button to a new row. It would add a more white-space/breathing room, so visually it doesn't feel so crammed.

I would like to handle this in a separate issue.
This separate menu row for the tag list could also be utilized in the context of #540.

tobias-klein:
> > #594 Tag statistic button in the toolbar can be disabled if there are no tags, otherwise it's very confusing.
> 
> I agree! The alternative would be to show a help message like for the dictionary panel and comparison panel.

@zhuiks Is there already an API for disabling the panel buttons? I did not find one.
Independent of this, there is also the usecase that you switch from a book to a search or tagged verse list and in that case an opened tag statistics panel would also become empty. Should we switch to the tag list in that case or maybe simply hide the side panel completely once a "non book verse list" has opened?

zhuiks:
> > > #594 Tag statistic button in the toolbar can be disabled if there are no tags, otherwise it's very confusing.
> > 
> > 
> > I agree! The alternative would be to show a help message like for the dictionary panel and comparison panel.
> 
> @zhuiks Is there already an API for disabling the panel buttons? I did not find one. Independent of this, there is also the usecase that you switch from a book to a search or tagged verse list and in that case an opened tag statistics panel would also become empty. Should we switch to the tag list in that case or maybe simply hide the side panel completely once a "non book verse list" has opened?

No, there is no any API for disabling panel buttons. I can work on this. As a part of this panel buttons behavior, the implementation can be that it will switch to the previous button/panel whenever button is set to disabled.

tobias-klein:
> No, there is no any API for disabling panel buttons. I can work on this. As a part of this panel buttons behavior, the implementation can be that it will switch to the previous button/panel whenever button is set to disabled.

Ok! For now I have added some logic in TagsController to reset the active panel when switching tabs or switching to a different text type in the current tab. You can find the logic in [`TagsController.resetActivePanelToTagPanel()`](https://github.com/ezra-bible-app/ezra-bible-app/blob/1cbe66765ecae3be0243704436545dedc8ba3b3a/app/frontend/controllers/tags_controller.js#L109).

tobias-klein:
@zhuiks
There was a significant deterioration of performance for switching from the tag list to other panels or hiding it - after your latest commits.

This is measured on my tablet and only captures the hiding of the tag list panel:
image

You see that re-calculate style takes more than five seconds.

This is how it still looked with 6240a315.
image

Here re-calculate style takes only ~40 ms.

I didn't analyze where this is coming from exactly. Maybe we can fix this with adjustments to the tag list related CSS classes and specifically the content-visibility attributes.

zhuiks:
Thanks for bringing that up, I will look into this.

tobias-klein:
@zhuiks 
The latest commit (mostly) mitigates the performance issue.

I added this:
```css
.checkbox-tag *:not(.cb-input-label-stats) {
  content-visibility: auto;
}
```

One thing that changed between the current version of the verse list and the previous one is that there is more nesting and the overall amount of elements per .checkbox-tag element increased.

Here is the latest measurement on Android:

image

zhuiks:
I don't have enough tags to replicate the issue :) But I assume the main cause was these kind of selectors:
```css
#tool-panel input:checked ~ label.button-small
```
I replaced ~ with + which might be faster.

> One thing that changed between the current version of the verse list and the previous one is that there is more nesting and the overall amount of elements per .checkbox-tag element increased.
The change above has the effect that the buttons are rendered mostly dynamically (you see that when scrolling fast in a long tag list), while the texts are "pre-rendered". I previously tried for all child nodes of .checkbox-tag, but that was too messy on Android.

Did you mean tag list?

I can't experiment much with it as my tag list is too short. Before you added `content-visibility: auto` have you had any performance issues when selecting the verse? (it would change the style of the tag button)

If the tag button is the root cause, the proper solution would be updating classes/attributes of the tag button directly instead of underlying checkbox attributes

tobias-klein:
> I don't have enough tags to replicate the issue :) But I assume the main cause was these kind of selectors:
> 
> ```css
> #tool-panel input:checked ~ label.button-small
> ```
> 
> I replaced ~ with + which might be faster.
> 
> > One thing that changed between the current version of the verse list and the previous one is that there is more nesting and the overall amount of elements per .checkbox-tag element increased.
> > The change above has the effect that the buttons are rendered mostly dynamically (you see that when scrolling fast in a long tag list), while the texts are "pre-rendered". I previously tried for all child nodes of .checkbox-tag, but that was too messy on Android.
> 
> Did you mean tag list?
> 
> I can't experiment much with it as my tag list is too short. Before you added `content-visibility: auto` have you had any performance issues when selecting the verse? (it would change the style of the tag button)
> 
> If the tag button is the root cause, the proper solution would be updating classes/attributes of the tag button directly instead of underlying checkbox attributes

Yes, I meant the tag list.
I could not see how the last change impacted the performance in any direction.

In general, loading the tag list does not feel as snappy anymore as compared to master. I think the root-cause is simply the number of nested dom elements in the list. I'll have another look in the next few days.

tobias-klein:
@zhuiks I created a script so that it's easier to reproduce issues with long tag lists. This currently only runs on Linux because of the static config directory setting in the script.

1) Important: Create a test user (or backup your ezra.sqlite)
2) Start the application once to initialize an empty database
3) Execute `npm run generate-long-tag-list`. You'll end up with a list of 1000 tags.

zhuiks:
@tobias-klein Cool, thanks!

tobias-klein:
@zhuiks The performance issues are fixed. I removed the element nesting in the tag list and changed the code so that it toggles the button using the `active` class. I still need to test this on Cordova.

zhuiks:
There is a layout bug (extra space) on large screen that was introduced by this PR:
Screenshot from 2022-01-08 13-57-24
Screenshot from 2022-01-08 13-58-11

zhuiks:
> > No, there is no any API for disabling panel buttons. I can work on this. As a part of this panel buttons behavior, the implementation can be that it will switch to the previous button/panel whenever button is set to disabled.
> 
> @zhuiks
> 
> Ok! For now I have added some logic in TagsController to reset the active panel when switching tabs or switching to a different text type in the current tab. You can find the logic in [`TagsController.resetActivePanelToTagPanel()`](https://github.com/ezra-bible-app/ezra-bible-app/blob/1cbe66765ecae3be0243704436545dedc8ba3b3a/app/frontend/controllers/tags_controller.js#L109).

I added the API to enable/disable a certain panel.

> As I was looking into this I had another thought. It may be useful to store the currently active panel per tab and restore it when switching to the respective tab. What do you think about that?

As for now, my thought is to not automatically switch panels on user. Although I just implemented automatic switch to previous button when button is disabled.

tobias-klein:
> > Ok! For now I have added some logic in TagsController to reset the active panel when switching tabs or switching to a different text type in the current tab. You can find the logic in [`TagsController.resetActivePanelToTagPanel()`](https://github.com/ezra-bible-app/ezra-bible-app/blob/1cbe66765ecae3be0243704436545dedc8ba3b3a/app/frontend/controllers/tags_controller.js#L109).
> 
> I added the API to enable/disable a certain panel and added the call to the tag-statistics-panel.
> 
> > As I was looking into this I had another thought. It may be useful to store the currently active panel per tab and restore it when switching to the respective tab. What do you think about that?
> 
> As for now, my thought is to not automatically switch panels on user. Although I just implemented automatic switch to previous button when button is disabled.

Thank you, Evgen! I'll resume the work soon.
Thanks also for reporting the layout bug - I have not seen that before.

zhuiks:
> There is a layout bug (extra space) on large screen that was introduced by this PR

So far I found out that it's triggered somewhere under this function [handleBookLoadingModeOptionChange](https://github.com/ezra-bible-app/ezra-bible-app/blob/5b3cdb573c5f16fa6eb0511ff7c0b00f443a9a13/app/frontend/components/options_menu/options_menu.js#L458).

@tobias-klein Can you reproduce this bug? Right now, I'm not sure if it's related to the current changes. But I can't reproduce it in master.

zhuiks (dismissed):

tobias-klein:
@zhuiks Thanks for your work to fix the layout bug. I still need to do some more testing/debugging on this PR. There is still some bug related to the update of the tag statistics.

tobias-klein:
Thanks @zhuiks For supporting with this. I corrected a few more bugs I found. I am merging this now ... remaining bugs will be fixed directly on `master`.

PR #597, opened 2021-12-30
Enhance tag selection menu: Support multi tag selection

### LINKED ISSUE #528: Easily choose many tags at the same time
Hi guys! When I want to explore some tags verses, I need to choose tags one at a time, since the options menu doesn't allow multiple marking. 

choose_tags

Is it possible to put a option to easily "mark all' tags? Or even allow the possibility to mark multiple tags at time?

Warm regards!

tobias-klein:
This fixes #528.

multi-tag-selection


tobias-klein:
@debritto This was a rather simple enhancement. Any feedback?

The recorded gif shows everything, but if you would like to test it ... here is the build:
https://github.com/ezra-bible-app/ezra-bible-app/suites/4792198148/artifacts/134130195

debritto:
Hi @tobias-klein 
Thank you for the implementation! This is exactly what I'm looking for. Just one more idea, what do you think about put a check box for "Select all"? In this way we can cover all use cases. ;)

tobias-klein:
> Thank you for the implementation! This is exactly what I'm looking for. Just one more idea, what do you think about put a check box for "Select all"? In this way we can cover all use cases. ;)

@debritto
Thanks for the feedback! I could add such a "Select all" button. However, I think this use case only makes sense if a user has only a small amount of tags in the database. How about adding that button, but only showing it if you have less than 50 tags?

debritto:
@tobias-klein !

Good idea! 
By the way, Happy New Year my Brother!

tobias-klein:
@debritto I just added "Select all" and "Deselect all" buttons. The "Select all" button will only be shown with 50 or less tags in the database.

@zhuiks This is ready for review now.

debritto:
Perfect!

tobias-klein:
@zhuiks Any feedback before I merge this PR?

zhuiks (approved):
Looks good to me!
Thank you.

Issue #352, opened 2021-06-17
URLs in SWORD Module Description are not clickable


Berovowilare:
Hello 

The URLs are not active links.

<img width="708" alt="Bildschirmfoto 2021-06-17 um 20 19 08" src="https://user-images.githubusercontent.com/36891575/122452201-4b9dfd80-cfa9-11eb-90d4-759f5a6bad56.png">

Thanks

smpurkis:
I've made [PR 609](https://github.com/ezra-bible-app/ezra-bible-app/pull/609) to try and solve this.

PR #609, opened 2022-01-22
Issue 352 - Make URLs in Sword Module Description Clickable

smpurkis:
Hello,

This PR is to fix [issue 352](https://github.com/ezra-bible-app/ezra-bible-app/issues/352).

I've new to this project, so am not sure if there is functionality for clickable URLs defined somewhere else in the codebase. If there is, would someone be able to point me to that functionality.

I continued my efforts assuming this functionality wasn't in the codebase already.
I've forked and made a branch "issue-352" that adds this functionality.

The code came from https://stackoverflow.com/a/1500501 and the Regex came from https://www.codegrepper.com/code-examples/whatever/use+regex+to+get+urls+from+string.

issue-352

I'm sure there are some changes that will need to be made.

### DIFF
--- a/app/frontend/components/info_popup.js
+++ b/app/frontend/components/info_popup.js
@@ -108,7 +108,7 @@ class InfoPopup {
     const swordPath = await ipcNsi.getSwordPath();
 
     const swordModuleHelper = require('../helpers/sword_module_helper.js');
-    var moduleDescription = await swordModuleHelper.getModuleDescription(currentBibleTranslationId);
+    const moduleDescription = await swordModuleHelper.getModuleDescription(currentBibleTranslationId);
     const moduleInfo = await swordModuleHelper.getModuleInfo(currentBibleTranslationId, false, false);
 
     const exportUserDataHint = await i18n.t('general.export-user-data-hint');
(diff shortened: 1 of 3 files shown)


smpurkis:
Accidently made "Merge branch 'master' into master", is there a quick way to fix this?

tobias-klein:
First of all, thank you very much @smpurkis!

I just tested this with the NASB module (from Lockman foundation) and in that case there are already properly marked up URLs in the module description. I think that's a special case that we need to handle somehow.

The text looks like this before being modified:
image

And when applying the changes from this PR, this is the result:
image

Which looks like this in the popup (invalid markup + doubled URL):
image

> Accidently made "Merge branch 'master' into master", is there a quick way to fix this?

We can later squash merge the changes of this PR and then this will disappear.

app/frontend/components/info_popup.js
```diff
@@ -121,6 +121,17 @@ class InfoPopup {
       `;
     }
 
+    function urlify(text) {
+      // replace urls in text with <a> html tag
+      // regex extracted from https://www.codegrepper.com/code-examples/whatever/use+regex+to+get+urls+from+string
+      var urlRegex = /(?:(?:https?|ftp|file):\/\/|www\.|ftp\.)(?:\([-A-Z0-9+&@#\/%=~_|$?!:,.]*\)|[-A-Z0-9+&@#\/%=~_|$?!:,.])*(?:\([-A-Z0-9+&@#\/%=~_|$?!:,.]*\)|[A-Z0-9+&@#\/%=~_|$])/igm
+      return text.replace(urlRegex, function(url) {
+        console.log(url)
+        return '<a href="' + url + '" target="_blank" title="' + url + '">' + url + '</a>';
```
tobias-klein, line 130: 2 things

* Remove the console.log - this is only for debugging
* I would suggest to use a [template literal](https://www.w3schools.com/js/js_string_templates.asp) for putting the variables into the markup.

tobias-klein (commented):
Looks fine in general. One special case is when there are already valid urls in the markup. In that case the conversion cannot be applied.

Style: Even though it's not required in JS we do ; after statements - and that's missing for line 127 and 133.

app/frontend/components/info_popup.js
smpurkis: - Have removed the console.log in the latest changes, ty for spotting that.
- The title attribute enables the URL to appear on hover over the link. I have removed it in latest changes. 
- Change made.

I wasn't sure the best place to put the "urlify" function, any suggestions.

smpurkis:
I've tried to account for the
```
<a href="https://www.google.com/">https://www.google.com/</a>
```
case.

Mind retrying it now?

tobias-klein:
Thanks for the updates @smpurkis!
It looks better now.

I have one more suggestion. Since the module description is also used in the installation assistant (see screenshot below), I would suggest to move the `urlify` function into `sword_module_helper.js` and call it directly in the context of `getModuleDescription`. Then the text will contain the links both in the installation assistant as well as in the info popup.

image

smpurkis:
Nice, I have made the changes you suggested. I moved the `urlify` function to the bottom of `sword_module_helper.js` file, so the `module.exports...` are readable and applied `urlify` to `getModuleDescription`. 

Thank you for the warm welcome @tobias-klein! I've been looking for some open-source Christian software to contribute to for a little while. I'm glad for any opportunity to contribute and hopefully help the Church, (even in some small quality of life ways). :)

tobias-klein:
@smpurkis Thank you once more! This is merged now. I just did some style fixes after merging and also added a block that ensures URLs are prefixed with http:// if that's missing.

PR #611, opened 2022-02-05
Pre-load section headers and allow navigation to sections from other chapters

### LINKED ISSUE #340: Chapter navigation: Load section headers for complete book upfront
A pre-condition is to fix the latest SWORD trunk issues on Windows, so that the new section header loading function is available there, too.

tobias-klein:
This fixes #340.

For testing you need to set the book loading mode to "Open all books chapter-wise".

There are updates in `package.json` so you need to run `npm install`.

### DIFF
--- a/app/backend/ipc/ipc_nsi_handler.js
+++ b/app/backend/ipc/ipc_nsi_handler.js
@@ -243,8 +243,31 @@ class IpcNsiHandler {
       return this._nsi.getBookIntroduction(moduleCode, bookCode);
     });
 
-    this._ipcMain.add('nsi_getBookHeaderList', (moduleCode, bookCode, withAbsoluteVerseNumbers) => {
-      return this._nsi.getBookHeaderList(moduleCode, bookCode, withAbsoluteVerseNumbers);
+    this._ipcMain.add('nsi_getBookHeaderList', (moduleCode, bookCode) => {
+      var bookTextJson = this._nsi.getBookText(moduleCode, bookCode);
+
+      var bookTextHtml = "<div>";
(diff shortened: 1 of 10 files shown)


tobias-klein:
Hi @zhuiks

Would you mind testing and reviewing this PR?

zhuiks (dismissed):
Very nice! 
All the code looks good, although I didn't dig in into logic much.

I noticed two inconsistencies when switching translations:
1. Switching translation doesn't always update navigation headers. And it seems to be "stacked" in a certain version (switching to RusSynodal and then back to NASB):
Screenshot from 2022-02-09 19-00-03

2. When I turn on "heading navigation" while in translation without headers (Ukrainian) and then switch to translation with headers (NASB), navigation headers remain blank:
Screenshot from 2022-02-09 18-48-25

Let me know if you can't reproduce these issues. I'll dig into it more

tobias-klein:
@zhuiks Thanks for testing this! I applied some fixes for the issues you observed. Would you mind testing once more?

zhuiks:
I'm still getting empty headers in navigation in the following (quite unlikely) use case:
1. Turn nav-headers off in any translation;
2. Switch to the translation with headers;
3. Turn nav-headers on.

**Result:** Blank headers in nav (similar to the second screenshot earlier).

tobias-klein:
> I'm still getting empty headers in navigation in the following (quite unlikely) use case:
> 
> 1. Turn nav-headers off in any translation;
> 2. Switch to the translation with headers;
> 3. Turn nav-headers on.
> 
> **Result:** Blank headers in nav (similar to the second screenshot earlier).

Thanks for reporting this! I will fix this before merging.

PR #779, opened 2022-10-17
Brazilian Portuguese Localization updated

debritto:
Updated to version 1.8


tobias-klein:
Thank you @debritto!

Would you mind providing another update to this PR, also considering these texts:

 * general
   * database-size
   * back
 * bible-browser
   * delete-note

PR #861, opened 2023-01-30
Reverting back to 'auto' instead of 'automático' in the Spanish translation to fix and close bug #809

### LINKED ISSUE #809: Language selection Auto - label growing too large for Spanish locale
With the Spanish locale, the label for the automatic language selection button grows to large, see the screenshot below.

All other languages have a label that works with less characters (like Auto in English).

image

reyespinosa1996:
I was searching the Internet and as long as a word have a very clear abbreviation and is commonly used in Spanish, it is okay to use it to translate that word. So I'm reverting the change back to Auto, so users with Spanish location set doesn't have interface issues. Hope this will close #809


tobias-klein:
Thank you @reyespinosa1996! Excellent.

PR #904, opened 2023-04-16
Update translation for Brazilian Portuguese

debritto:
Updated


tobias-klein (commented):
Thank you @debritto!

There is one more text that needs a translation:

`module-last-update`

Would you mind adding that?
After that I'll merge.

PR #924, opened 2023-05-22
Order languages in one list

### LINKED ISSUE #923: Long list of 'other' languages is rendered in several groups
Actual behavior:
The list of 'other' languages in the install assistant consists of several groups where each group is sorted alphabetically. This does not seem to make sense even though the groups are logical from a programming point of view (iso6391 languages, iso6392T languages, iso6393 languages).

Expected behavior:
One list of 'other' languages that is sorted alphabetically in one block.

tobias-klein:
This fixes #923.

### DIFF
--- a/app/frontend/components/module_assistant/assistant_helper.js
+++ b/app/frontend/components/module_assistant/assistant_helper.js
@@ -129,7 +129,7 @@ function generateCheckbox(item, checked, options) {
   } else {
     const { code, text, description, count, disabled, ...rest } = item;
 
-    const style = text && text.length > 22 && (options.columns === 'auto-fill' || options.columns > 1) ? 'style="grid-column-end: span 2"' : '';
+    const style = text && text.trim().length > 22 && (options.columns === 'auto-fill' || options.columns > 1) ? 'style="grid-column-end: span 2"' : '';
 
     const extraAttr = Object.entries(rest).map(([attr, val]) => `${attr}="${val}"`);
 
@@ -144,7 +144,7 @@ function generateCheckbox(item, checked, options) {
         ${description ? `description="${description}"` : ''}
         ${options.info ? 'flex-align-top' : ''}
         ${extraAttr.join(' ')}>
-        ${text ? text : code}
+        ${text ? text.trim() : code}
       </assistant-checkbox>`;
 
     return checkbox;
--- a/app/frontend/components/module_assistant/step_languages.js
+++ b/app/frontend/components/module_assistant/step_languages.js
@@ -184,6 +184,7 @@ class StepLanguages extends HTMLElement {
     await waitUntilIdle();
     const containerLongList = document.createElement('div');
 
+    // Merge several groups of languages into one set of 'other' languages
     languages['other-languages'] = new Map([...languages['iso6391-languages'], ...languages['iso6392T-languages'], ...languages['iso6393-languages']]);
 
     for(const category of ['other-languages', 'unknown-languages']) {      
--- a/app/frontend/components/module_assistant/step_languages.js
+++ b/app/frontend/components/module_assistant/step_languages.js
@@ -183,11 +183,14 @@ class StepLanguages extends HTMLElement {
       
     await waitUntilIdle();
     const containerLongList = document.createElement('div');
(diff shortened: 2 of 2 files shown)


tobias-klein:
@zhuiks Do you think this change makes sense? Was there a particular reason to have the languages grouped in the categories of iso6391, iso6392T, iso6393?

zhuiks:
> @zhuiks Do you think this change makes sense? Was there a particular reason to have the languages grouped in the categories of iso6391, iso6392T, iso6393?
@tobias-klein from what I can recover from my memory and a little bit of digging in the code different iso standards have different value of language details and language/region name localization. If you switch locale to something that is not using Latin characters (i.e. Ukrainian or Russian), the last iso group or maybe last two would not have localized names. 

Also iso groups would represent languages by number of speakers (more/less used languages). As far as I know for most iso6393 languages that don't have any representation in iso6391 or iso6392T speakers would be bilingual or trilingual with the languages from more "common" groups. 
It's much easier to find language in the smaller group than in the one big "merged" group. (E.g. compare your UX scrolling down to Tatar or Turkish or Uzbek with merged group and groups divided by number of speakers) 

I hope that what I wrote make sense. Let me know what you think.

tobias-klein:
@zhuiks I understand your point. However, this is not self-explanatory to the user, is it? When you scroll through and see that the sorting suddenly starts again and then again, you don't know why it is like that and what each block of languages entails. If we keep it like this, we should either add an explanation above the language list or maybe additional headlines over each block, or maybe both?!

zhuiks:
@tobias-klein yes, I agree! I just don't know how to phrase it. 
[This article](https://lingohub.com/blog/2019/04/ultimate-guide-iso-language-codes) seems to be a good background on different iso 639 standards
> [ISO 693-1](https://en.wikipedia.org/wiki/ISO_639-1) ...There are 183 two-letter codes registered as of June 2021. The registered codes cover the world's major languages

So maybe instead of `iso6391-languages` we can use `common-languages` with heading "Common Languages"

> The ISO 639-2 firstly includes identifiers for languages from the 639-1 and additional languages which have a relevant amount of literature. This list provides identifiers for language families, and so covers almost all languages of the world...

Instead `iso6392T-languages` use `less-common-languages` with heading "Less Common Languages"
As I've stated above language names from these two groups can be translated to locales that we have based on [Internationalization API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames/of) (which I think derives its data from commonly used [Unicode CLDR project](https://cldr.unicode.org/index)). I think total of languages in these groups should be around 450.

> The ISO 639-3 list covers individual languages from ISO 639-2, but includes also extinct, ancient, historic, and constructed languages (conlangs). This language code set recognizes the lesser-known languages.

Instead `iso6393-languages` use `rear-languages` with heading "Rear Languages" This group has close to 2K languages.

`unknown-languages` are "Non-Standard Language Codes". These codes do not have equivalent in ISO 639-3 standard.

The language data is taken from [`languages.json`](https://github.com/ezra-bible-app/ezra-bible-app/blob/master/lib/languages.json) which is a compilation of all SWORD locales with extra info from Internationalization API and ISO 639-3 auto-generated by [`sword_locales_to_json.js`](https://github.com/ezra-bible-app/ezra-bible-app/blob/master/build_scripts/frontend/sword_locales_to_json.js#L45).

I can put together a PR with these proposed headings above each language group and add a paragraph with a bit of explanation.

I don't mind making less groups. I just think that having ~450 common languages separate from ~2000 rear languages is a better UX.

tobias-klein:
@zhuiks Thanks for your feedback. I understand the reasoning behind limiting the number of entries in the groups shown. Using an approach that enhances the usability is good. We should definitely make the logic for the language groups transparent to the user, though.

However, before moving forward with implementation it may be interesting to put this up for discussion in the SWORD mailing list and get some feedback there about the idea with the groups and the headers. What do you think?

When you say "Rear languages" do you actually mean "Rare languages"?

zhuiks:
> However, before moving forward with implementation it may be interesting to put this up for discussion in the SWORD mailing list and get some feedback there about the idea with the groups and the headers. What do you think?
@tobias-klein sound like a great idea. Do you mind communicating that?

> When you say "Rear languages" do you actually mean "Rare languages"?
:grin: yes!

Issue #944, opened 2023-11-04
Locale update for 1.12.0


tobias-klein:
There is one new text for 1.12.0. This is related to #921.

* bible-browser.adjust-font-size-side-panel

Translation status:

* [x] Brazilian Portuguese
* [x] Dutch
* [x] French
* [x] German
* [x] Romanian
* [x] Russian
* [x] Slovakian
* [x] Spanish
* [x] Ukrainian

@lemtom Could you submit a pull request for the French and Dutch translation.
@MartinIIOT Could you submit a pull request for the Slovakian translation.
@reyespinosa1996 Could you submit a pull request for the Spanish translation.
@augustin-colesnic Could you submit a pull request for the Romanian translation.
@debritto Could you submit a pull request for the Brazilian Portuguese translation.
@zhuiks Could you submit pull requests for the Ukrainian and Russian translation.

I am planning to release 1.12.0 on **November 18**. Thanks in advance for your help!

MartinIIoT:
#945 Job Done.

augustin-colesnic:
Hi @tobias-klein ,
here is the PR #947

reyespinosa1996:
Check #950

tobias-klein:
@debritto @zhuiks Friendly reminder :)

tobias-klein:
Thank you all! Right on time! :+1:

PR #947, opened 2023-11-09
i18n(romanian): add translation for bible-browser item

augustin-colesnic:
as per issue #944


tobias-klein (approved):
Thank you very much @augustin-colesnic!

PR #951, opened 2023-11-13
Update translation.json

debritto:
Brazilian Portuguese - panel font size - adjusted.


tobias-klein (approved):
Thank you very much @debritto!

PR #1010, opened 2024-01-17
Si slovenian translation

msavli:


tobias-klein:
This PR is not needed anymore, instead I am going to integrate #1011.

Issue #1047, opened 2024-06-16
Locale update for 1.14.0


tobias-klein:
These are the updated texts for [1.14.0](https://github.com/orgs/ezra-bible-app/projects/6/views/1).

* New section `type-face`
see #1007
* `tags.tag`
see #1012
* `commentary-panel.copy-commentary-to-clipboard` and `commentary-panel.copy-commentary-to-clipboard-success`
see #1041

Translation status:
* [ ] Brazilian Portuguese
* [x] Dutch
* [x] French
* [x] German
* [x] Romanian
* [x] Russian
* [x] Slovakian
* [x] Slovenian
* [x] Spanish
* [x] Ukrainian

@lemtom Could you submit a pull request for the French and Dutch translation.
@MartinIIoT Could you submit a pull request for the Slovakian translation.
@reyespinosa1996 Could you submit a pull request for the Spanish translation.
@augustin-colesnic Could you submit a pull request for the Romanian translation.
@debritto Could you submit a pull request for the Brazilian Portuguese translation and also address #1013.
@zhuiks Could you submit pull requests for the Ukrainian and Russian translation.
@msavli Could you submit a pull request for the Slovenian translation.

I am planning to release 1.14.0 on June 30. Thanks in advance for your help!

tobias-klein:
@msavli @debritto @zhuiks
Will trigger the release process tomorrow. It would be great to still get all the locale updates in. Thanks in advance!

PR #1055, opened 2024-06-23
Spanish translation for v1.14 adresses #1047

reyespinosa1996:
I little bit late but translations into Spanish are in place for the new release.


tobias-klein (approved):
Thank you very much @reyespinosa1996! :)

PR #1059, opened 2024-06-24
[locale]-add romanian locale update for 1.14.0

augustin-colesnic:
as per #1047


tobias-klein (approved):
Thank you very much, Augustin! :)

PR #1062, opened 2024-06-30
Slovenian local update

msavli:


tobias-klein (approved):
Thank you very much @msavli! Just in time :)

PR #1070, opened 2024-07-03
Enable Text Selection in Commentaries

mrnazu:
This PR adds CSS property (user-select: text;) to the #commentary-panel-content section in commentary_panel.css. By enabling text selection within the commentaries, users can now highlight and copy specific passages. This enhancement improves usability and aligns with common user expectations.

For more info, check this out: https://github.com/ezra-bible-app/ezra-bible-app/discussions/1069


tobias-klein:
Hi @mrnazu! Thank you for your suggestions in #1070 and this pull request.
THis is a good idea!

However, to enable copying of text from the text selection in the Commentary panel, we need more than just the CSS change. We need to implement a handling either via shortcut (CTRL+c) or via a context menu, so that the selected text can be copied to the clippboard. Aside from that, we also need to consider the behavior on Android.

What could be done is the following:
1) Add a button "Copy selected commentary text to clipboard". That button would only be enabled/functional once text has been selected.
2) Detect the selection of the text. Enable the button if there is a selection and disable it if nothing is selected.
3) If there is a selection and the button is clicked, copy the selected text to the clipboard (there are existing functions for this both on Electron as well as Cordova/Android).

All of this should probably be implemented in the Commentary Panel code in `frontend/components/tool_panel/commentary_panel.js`.

Your thoughts?

mrnazu:
> Your thoughts?

Sure, will do that. Thanks!

mrnazu:
Your thoughts?

tobias-klein:
Hi @mrnazu

There seems to be something missing.

image

How did you test this?

tobias-klein:
Hi @mrnazu
How is it going? Any plans to finish this?

mrnazu:
> How is it going? Any plans to finish this?

Hey @tobias-klein 

I appreciate your patience. I've been quite busy recently, but I did my best to work on the feature. the copy-to-clipboard functionality has been implemented, and the corresponding updates have been made to the locale files. 

I was a bit delayed due to other commitments, but I’ve now committed and pushed the changes. Thank you for understanding!

package.json
tobias-klein, line 1: Unnecessary change - not needed in the context of this PR.

tobias-klein (changes requested):
@mrnazu 
Thank you very much for adding the functionality. I tested briefly and the copy action based on selected text works! Congrats! Good job :)

When looking at your code changes I noticed the following:
- You changed quite a lot of code style in commentary_panel.js - those are not related to the requested functionality. Please undo those changes. We may discuss and change separately. But let's separate concerns.
- You added a "custom-notes-panel" in tool_panel.html. Please undo - it is not related to the copy commentary functionality, but may be discussed/changed separately.
- When testing I noticed that the button text was empty. Then I checked the locales and did not find the `commentary-panel.copy-button` text. It may be missing still in what you pushed?
- I would suggest to disable/enable the button when text is selected vs. when no text is selected.
- I would suggest to align the margin-left of the button so that it is aligned with the headlines of the commentaries. Currently it is too far to the left. We should also add some margin-top, as the button appears squished right underneith the top of the box.

mrnazu:
That's very awesome and thank you for the feedback! actually, the code style issue was a mistake on my part. I do this daily after finishing my tasks: I right-click and refactor using the Prettier code formatter. 

I've made the following updates based on your suggestions:

- I’ve restored the previous code style in commentary_panel.js and removed unrelated style changes. (I tried by best)
- The "custom-notes-panel" added to tool_panel.html has been removed as it was not related to the copy commentary functionality. we can discuss any changes to it separately if needed.
- I've added the missing "commentary-panel.copy-button" text for all specified languages using the keyword "copy-button." However, I've now replaced every instance of the keyword "copy-button" with "commentary-panel.copy-button." The button text should now appear correctly in all locales.
image
- The button is now enabled/disabled based on text selection. This is accomplished by adding/removing the "ui-state-disabled" class.
image
- Adjustments have been made to the margin-left and margin-top of the button to align it properly with the commentary headlines and avoid it appearing squished.
image
image


Please let me know if there are any further adjustments needed. Thanks again for your thorough review!

tobias-klein:
Hi @mrnazu!
Thank you once more for your work! This is merged now. I applied a few more changes for integration, consider my latest commits on the master branch.
The functionality will be released as part of the next version 1.15.

mrnazu:
Thanks a lot and God bless :))

PR #1083, opened 2024-07-24
feat: commentary selection handling for ctrl+c shortcut

mrnazu:
## shortcut functionality in the `AppController`
 **Changes:**
 - Added `dictionarySelectionActive` property to track dictionary selection state.
 
image

 - Modified `initGlobalShortCuts` to copy dictionary text to clipboard when dictionary selection is active.
 
image

I hope, this update ensures that the correct text is copied based on the current selection state.


tobias-klein:
Hi @mrnazu!

Thank you very much for your contribution!

I did not have a chance to test this, yet. However, I took look at the code and noticed two things:

- You are not calling the new function `setDictionarySelectionActive()` anywhere. Did you test your change successfully? Is that function actually needed?
- You used 4 spaces for indentation instead of 2. The standard for JavaScript code is 2 and that's what is used for the rest of Ezra Bible App as well. Also: You modified areas of the code that are not related to the purpose of this PR, like the Mousetrap blocks for all the other shortcuts.

app/frontend/controllers/app_controller.js
```diff
@@ -250,61 +259,65 @@ class AppController {
...
     if (platformHelper.isMac()) {
-      shortCut = 'command+c';
+        shortCut = 'command+c';
     }
 
     Mousetrap.bind(shortCut, async () => {
-      await this.verse_selection.copySelectedVerseTextToClipboard();
-      return false;
+        if (this.dictionary_controller.isSelectionActive()) {
+            await this.dictionary_controller.copySelectedTextToClipboard();
```
tobias-klein, line 267: The function `copySelectedTextToClipboard()` is not async and therefore does not need to be called with `await`.

mrnazu: Done. Removed await keyword from verse_selection.copySelectedVerseTextToClipboard() call

mrnazu:
Hey :)
there is an update..
- I added the `isSelectionActive` method just below the `setDictionarySelectionActive` method inside the `AppController` class.
- And Inside the `initGlobalShortCuts` method, I replaced the usage of `this.dictionary_controller.isSelectionActive()` with `this.isSelectionActive()` to utilize the new method correctly.

tobias-klein:
Hi @mrnazu,

I tested this and it seems to work. However, I do not know why, because we are actually interacting with the wrong component (dictionary controller instead of app_controller.commentaryPanel) and there are currently Javascript Errors when invoking ctrl+c.

image

So why does the text still reach the clipboard?
My guess is, because we are not calling `this.verse_selection.copySelectedVerseTextToClipboard()` and therefore the clipboard does not get overwritten. It seems like anyway the Browser is already copying the text to the clipboard unless we overwrite it with something else.

So, based on that you need to do the following things now:
1. Call the right function in the shortcut handler: It should be app_controller.commentaryPanel.copySelectedTextToClipboard()
1. The comment about the indentation and unnecessary code changes must still be addressed (see above).
1. The Member variable AppController.isDictionarySelectionActive and the respective getter/setter functions need to be removed, because they are not used. Also, this is not about dictionary selection, but commentary selection! I commented wrongly in the discussion at first (saying "dictionary selection" instead of "commentary selection"), so that was my bad.

Some general hints for all Pull Requests:
- Please make sure that there are no JavaScript errors. You can check this in the developer tools console that shows in the development build when pressing ctrl+shift+i.
- Make sure that all code is actually used.
- Do not apply custom code formatters and only modify the parts of the code that need to be modified based on the summary of the PR.

mrnazu:
Hi
I've encountered several internet issues while trying to install dependencies for this PR and the our last PR. I haven't been able to test the code properly. Due to starting a new project, I won't be able to continue working on this. Could you please finish the PR and release the next version? Thanks!

tobias-klein:
Closed due to inactivity.

PR #1108, opened 2024-09-14
Si slovenian translation1

msavli:


tobias-klein:
Hi @msavli 
It seems like you have used an old local branch to create this PR.
The latest changes from master are not in here.

We can do two things now:
1) Either we merge the changes from master and then make the locale updates.
2) Or you re-create a new branch based on the latest master and do the locale updates there.

Option 2) may be easier.

How would you like to proceed?

PR #1116, opened 2024-09-24
Update Slovenian locale

msavli:


tobias-klein (approved):
Thank you very much @msavli!
Excellent.

PR #1146, opened 2024-11-04
Update electron -> 28.0.0 to improve note taking under Wayland

ormris:
Using Ezra on Linux using the Wayland windowing system creates lag and stuttering when typing notes. See this video (I'm typing at a constant rate):

[Screencast from 2024-11-03 17-42-00.webm](https://github.com/user-attachments/assets/d47b512e-826c-4e63-af68-23c9023bb978)

Unfortunately, this is due to rendering issues in Chromium rather than configuration issues in the app. In my testing, Electron v28 was the oldest version that supports Wayland in a usable way. This is not only due to rendering improvements (this is the first version without stutter), but also support for window decorations in Wayland (see: https://releases.electronjs.org/release/v28.0.0). I ran the test suite before and after upgrading electron on Linux and it produced the same results when running in X11. However, the test suite does not currently run on Wayland. My guess is that this has something to do with screenshots which are handled very differently in the two systems. I realize there may be other changes needed related to such a large jump, so let me know if there's any other work needed. Everything worked well for me.

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,7 +115,7 @@
     "chai-as-promised": "^7.1.1",
     "copy-dir": "^1.3.0",
     "cucumber": "^6.0.5",
-    "electron": "^17.1.0",
+    "electron": "^28.0.0",
     "@electron/osx-sign": "^1.0.5",
     "electron-packager": "^17.0.0",
     "@electron/rebuild": "^3.6.0",


tobias-klein:
@ormris I knew that there would be eventually a good reason for upgrading Electron, despite the fact that the test framework would not work anymore. The test framework uses Spectron and the last [Electron version supported by Spectron is 17.0.0](https://github.com/electron-userland/spectron?tab=readme-ov-file#version-map). That is the reason why we are currently stuck with Electron 17.

There are more things in Ezra that may be incompatible with newer versions of Electron. I believe there are also some changes related to the [electron/remote](https://github.com/electron/remote) package that require some code adjustments. I am not sure about the exact impact on functionality.

At the end it is possible fairly easily to migrate to a new Electron version if you ignore the test framework. What we could do for now (to prioritize desktop compatibility over test framework compatibility) is to only use the old Electron version during testing. That would enable compatibility with Wayland on the desktop and ensure that the general functionality is still regression-tested. The majority of the functionality in Ezra does not depend on specific Electron features. The proof of that is that it also runs in the Chromium wrapper provided by the Cordova framework on Android.

If we upgrade Electron, we may as well upgrade to the latest stable version, though. That is currently 33.x.

The GitHub test workflow is defined in `.github/workflows/test.yml`. There we could simply add another step after `npm install` to install a specific version of Electron before running the tests.

So, I am ok to move forward with this based on these steps:
1) Use the latest Electron stable.
2) We may need to upgrade some more dependencies, especially the ones with Electron in the package name.
3) Add a step in the test workflow to keep using Electron 17.1.0 in the test suite.
4) Perform extensive manual testing of Ezra with the latest Electron version. The testing should also be done on Windows and macOS. I could help with that.
5) Check for additional code changes needed by the upgrade (e.g. electron/remote).

ormris:
Sounds good. I'll take a look.

tobias-klein:
> Task 1) and 3) are pushed. 2) is a bit more difficult. I would love to avoid iterating between 2) <-> 4), updating one dependency at a time. Is there another way? I'll look at 5).

Thank you!

I just tested this locally.

When I ran the app based on this branch I had JavaScript errors related to @electron/remote. I then updated @electron/remote to the latest version and the JavaScript errors disappeared.
So please already put that one to the latest version 2.1.2. I will now proceed with other tests.

tobias-klein:
@ormris
There are more places where the Electron version needs to be changed. See this list:

image

ormris:
All comments should be addressed.

>The app seems to work with the new Electron version once @electron/remote has been updated to 2.1.2. I still need to check and test how things work on macOS and Windows. I'll do that later this week.

That's good to hear. I wasn't able to find any documentation for additional code changes.

tobias-klein:
> All comments should be addressed.
> 
> > The app seems to work with the new Electron version once @electron/remote has been updated to 2.1.2. I still need to check and test how things work on macOS and Windows. I'll do that later this week.
> 
> That's good to hear. I wasn't able to find any documentation for additional code changes.

Thanks!

I noticed that `npm run rebuild` failed on this branch.
I had to add the following dependency under dev dependencies to make it work again:

`"node-abi": "^3.71.0"`

So please add that as well on your branch.

Also, the minimum GCC version now seems to be gcc-10. I need to consider that for the auto build system for some of the older distros. I'll take care of that later this week.

tobias-klein:
@ormris 
I am going to merge this to a development branch (`electron-update`) for now and then keep working on it there. Before merging it to master I need a few more tests related to GitHub Actions and I can only test this when merged.

tobias-klein:
@ormris The work will be continued under #1147. Thank you very much for your contribution!

ormris:
Thanks! Glad this is in.

Issue #1210, opened 2025-02-22
Search Function Unreliable


rgbprg3:
I love this app, but it appears the search function is a bit broken. Many times it fails to find certain phrases. For a practical example, searching for the phrase (without quotes) "generation to generation" will return results in Isaiah, Jeremiah, and Daniel, but will ignore this exact phrase which also appears in Exodus 17:16. Additionally, searching for just a single word such as "Amalek" will not return the word, clearly also in the same verse. 

I seem to recall other Bible apps such as Xiphos also suffer from this same broken search feature, which is sad. I hope it can be fixed in this app because I really like it. (Side note: E-Sword's search feature does find it.)

![Image](https://github.com/user-attachments/assets/2caf13ce-6276-49f0-850b-267183433816)

tobias-klein:
@rgbprg3 Thank you for reporting this. It seems to be an issue with the *exact phrase search* of the SWORD engine. Ezra Bible App and some other programs are relying on the SWORD engine. I will report this issue on the SWORD mailing list and see what they say there.

rgbprg3:
That's terrific news.

I'd be happy to help test a beta to ensure the search is working.
________________________________

tobias-klein:
@rgbprg3 This particular bug is fixed in the PR #1211. Would you like to test based on the GitHub Actions build that is available through that PR? I invited you as collaborator in this report. You should have access to the build results.

rgbprg3:
I'll be honest I have no idea what I'm doing, and I'm terrified I will mess something up. I have very limited knowledge of git and github especially (BAD, I know).

So, unless there is a completed binary I can put into a VM and do actual tests manually, I'm not going to be much help... I know I said I would help test, and I certainly want to, but I don't want to hold this up while I figure out what I'm doing.

I really want to learn git, but my current work schedule has not allowed me much free time.

Apologies.
________________________________

tobias-klein:
> I'll be honest I have no idea what I'm doing, and I'm terrified I will mess something up. I have very limited knowledge of git and github especially (BAD, I know).
> 
> So, unless there is a completed binary I can put into a VM and do actual tests manually, I'm not going to be much help... I know I said I would help test, and I certainly want to, but I don't want to hold this up while I figure out what I'm doing.
> 
> I really want to learn git, but my current work schedule has not allowed me much free time.
> 
> Apologies.
> […](#)

Hi @rgbprg3 
No worries.
Just check this page:
[...]

PR #1211, opened 2025-02-25
Increment node-sword-interface to version 1.0.37 to fix unreliable search

tobias-klein:
This fixes the unreliable search function reported in #1210.


tobias-klein:
@rgbprg3 Could you test one of the builds produced in this PR?

PR #1235, opened 2025-04-14
Slovenian translation

msavli:


msavli (commented):
Slovenian translation

PR #1333, opened 2026-02-21
Fix scroll position lost when switching between tabs

sse1234:
The scroll position of a tab could be lost (jumping to the beginning of the book) when switching between tabs. Three related issues were fixed:

- Replace window.location hash navigation with element.scrollIntoView() in restoreScrollPosition. The hash approach is global and fragile: concurrent async show handlers race to set the same hash, and setting the same hash value twice causes no scroll at all.

- Preserve existing saved scroll position when getFirstVisibleVerseAnchor returns null during tab switching (verse list temporarily hidden by loading indicator), instead of overwriting it with "top".

- Guard handleScrollEvent against stale debounced scroll events from non-visible tabs, which could save the wrong tab's verse anchor. Also clean up old scroll listeners before attaching new ones to prevent accumulation when text is reloaded.

Disclaimer: Vibe coded using Claude Opus 4.6

### DIFF
--- a/app/frontend/controllers/tab_controller.js
+++ b/app/frontend/controllers/tab_controller.js
@@ -592,14 +592,18 @@ class TabController {
     if (tabIndex != null && tabIndex >= this.metaTabs.length) {
       return;
     }
-    
+
     var metaTab = this.getTab(tabIndex);
     var firstVerseListAnchor = verseListController.getFirstVisibleVerseAnchor();
 
     if (metaTab != null) {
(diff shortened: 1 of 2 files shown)


tobias-klein:
@sse1234 Thanks for your contribution!

Would you mind outlining a few scenarios in the PR summary that describe how to reproduce the various issues?

Alternatively you could raise an issue and put those scenario descriptions there and then we can link that issue here.

I'll gladly review this, but I'd like to understand which exact scenarios you were testing with.

PR #1367, opened 2026-04-12
Security: Insufficient HTML sanitization before Markdown rendering enables XSS

tomaioo:
## Summary

Security: Insufficient HTML sanitization before Markdown rendering enables XSS

## Problem

**Severity**: `High` | **File**: `app/frontend/helpers/notes_helper.js:L32`

The note rendering pipeline removes only <script> and <style> tags, then passes the result to `marked.parse()`. This does not remove dangerous HTML attributes (e.g., `onerror`), `javascript:` URLs, embedded SVG payloads, or other active content. If note content is user-controlled or synced from external sources, this can lead to stored/DOM XSS when rendered.

## Solution

Use a robust HTML sanitizer (e.g., DOMPurify with a strict allowlist) on the final rendered HTML output, or disable raw HTML in Markdown parsing. Ensure dangerous URI schemes and event-handler attributes are stripped.

## Changes

- `app/frontend/helpers/notes_helper.js` (modified)

### DIFF
--- a/app/frontend/helpers/notes_helper.js
+++ b/app/frontend/helpers/notes_helper.js
@@ -30,29 +30,37 @@ module.exports.getTooltipText = function(noteText) {
 };
 
 function sanitizeNotes(input) {
-  // Create a DOM parser
   const parser = new DOMParser();
-
-  // Parse the input as an HTML document
   const doc = parser.parseFromString(input, 'text/html');
-  
-  // Remove all <script> elements
-  doc.querySelectorAll('script').forEach(script => script.remove());
-  
-  // Remove all <style> elements
-  doc.querySelectorAll('style').forEach(style => style.remove());
-  
-  // Return the sanitized HTML as a string
+
+  doc.querySelectorAll('script, style, iframe, object, embed, svg, math').forEach(element => element.remove());
+
+  doc.querySelectorAll('*').forEach(element => {
+    [...element.attributes].forEach(attribute => {
+      const name = attribute.name.toLowerCase();
+      const value = attribute.value.trim().toLowerCase();
+
+      if(name.startsWith('on') || name === 'style' || name === 'srcdoc') {
+        element.removeAttribute(attribute.name);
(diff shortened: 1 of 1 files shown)


tobias-klein:
Hi @tomaioo!
Under which realistic conditions is this an actual threat?

tomaioo:
Great question — this becomes a real threat whenever note text is not fully trusted (e.g., regular users, imported/synced notes from integrations, API clients, or compromised third-party systems).  
Because `marked.parse()` allows raw HTML and our current filter only removes `<script>/<style>`, payloads like `` or `javascript:` links can persist and execute when any user opens the note (stored XSS).  
In practice, that can lead to session/token theft or privileged actions if a maintainer/admin views the poisoned note.

tomaioo:
Great question — this becomes a real threat whenever note text is not fully trusted (e.g., regular users, imported/synced notes from integrations, API clients, or compromised third-party systems).  
Because `marked.parse()` allows raw HTML and our current filter only removes `<script>/<style>`, payloads like `` or `javascript:` links can persist and execute when any user opens the note (stored XSS).  
In practice, that can lead to session/token theft or privileged actions if a maintainer/admin views the poisoned note.

tomaioo:
Great question — this becomes a real threat whenever note text is not fully trusted (e.g., regular users, imported/synced notes from integrations, API clients, or compromised third-party systems).  
Because `marked.parse()` allows raw HTML and our current filter only removes `<script>/<style>`, payloads like `` or `javascript:` links can persist and execute when any user opens the note (stored XSS).  
In practice, that can lead to session/token theft or privileged actions if a maintainer/admin views the poisoned note.

tobias-klein:
Thanks for getting back. This is not a web-based tool and therefore there cannot be any session/token theft. There is also not a multi user system in Ezra Bible App with different levels of permissions (like maintainers/admins). I do not see a good enough reason for this change. Thanks for understanding.

PR #1378, opened 2026-04-25
Slovenian translation

msavli:


tobias-klein (approved):
Thank you so much @msavli

